PR #98, opened 2015-02-16
remove (indigo USE_DEB=true) travis test from allow_failures

wkentaro:
indigo用にテスト通過基準を変更しました。


jiangjun0105:
What's the meaning of this change? 
If it is not that much difficullt to explain, can you tell me?
Of course when you are free.

wkentaro:
This change is to allow failure during installation and catkin_make
building in indigo ROS envionment test on travis. "allow failure" means the
test is taken as succeeded even if it fails in the condition. There is
several conditions (ex. groovy, hydro.)

2015年3月1日日曜日、Jiang Jun<notifications@github.com
<javascript:_e(%7B%7D,'cvml','notifications@github.com');>>さんは書きました:

> What's the meaning of this change?
> If it is not that much difficullt to explain, can you tell me?
> Of course when you are free.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/start-jsk/2014-semi/pull/98#issuecomment-76562574>.
>


-- 
和田 健太郎 / Kentaro Wada
http://wkentaro.com <http://www.wkentaro.com>

PR #203, opened 2015-04-12
Update Baxter urdf model for long vacuum hand

aginika:
Change the length of the colisiion
vacuum_hand


wkentaro:
Do I have to prepare updated dae model?

aginika:
I got those things from Sugiura. So don't worry about it. I will updated it in another pull request

wkentaro:
Thanks!

PR #221, opened 2015-04-22
Update README to add How to execute DEMO

aginika:
Add how to execute demo.


wkentaro:
We should add this kind of document.
Thank you.

PR #233, opened 2015-04-23
Make move for verification as service

wkentaro:


k-okada:
Please consider writing with action lib
Since service call has chance to dead lock and If we working on autommous service, this is critical, for example if server or client segfaults we can not recover from that.
On the other hand, if we use action lib and start mode with respwan=true, then the service restarts and it has chance to continue working. It is also important to use parameter server to preserve current state so that respaened node can back to previous status.

PR #244, opened 2015-04-25
add Simulator Mode

aginika:
@wkentaro I changed someparts.
- We need to type `(apc-init)` and `(main)` . If you want to execute without typing every time, please type as below
```
roseus robot-main.l "(progn (apc-init)(main))"
```
and If you want to execute in launch, automatically, please write like as below link

https://github.com/jsk-ros-pkg/jsk_demos/blob/master/jsk_2013_04_pr2_610/launch/demo_common.launch#L17

- I added simulater mode , below parts check the `*ri*` is simulator mode or not. You can access with `*simulator-p*` as global variables.
  - With this, in pick-objects, skipping the grap check loop.
```
(setq *simulator-p* (not (send *ri* :joint-action-enable)))
```


aginika:
Please merge

aginika:
Please merge

PR #249, opened 2015-04-25
add some voices for notify

aginika:
For debug, it might be a little bit useful when debugging.

```
(speak-en "aaa" :google t)
```
will non blocking speaking

and 

```
(speak-en "bbb" :google t :wait t)
```
will blocking speaking. but this seems to have a little bug... I will check.


aginika:
please merge

aginika:
Could you merge this one too?
Or is there something wrong arount this my PR?

PR #270, opened 2015-04-27
Add robot main state machine and arm action server

aginika:
This is challenging to use smach to control dual arm.
Please still use robot-main.l as main controller.

For using the functions with 2arms
I added optional arm controller arguments to each function so that I can select the arm-controller.
So, please be careful about this change
apc_smach


k-okada:
nice, before we merge this one, Iet's design test case and check if this
works well.

= Trial 1 -- without recognition/verification =

- setup 12 objects, each bin have 1 object
- (try) to pick 12 object
- measure total time
- measure number of object picked

= Trial 2 -- with recognition =

- setup 25? objects
- (try) to pick 12 object
- measure total time
- measure number of object picked


--
◉ Kei Okada

aginika:
Ok, thanks for your advise.
I will try to use it.

k-okada:
also we'd like to count "score" and compere with highest possible score

--
◉ Kei Okada

aginika:
To make sure what I implement here,
this is just a state machine and doesn't use PDDL. All diagram above picture is written by myself.
This state machine is  to command each arm asynchronously.

In premise, above, I will implement below.
- count the `score` of total picking
- count the `time` of total picking
- store which objects I picked

I will claim here "please merge" when I implement all these things and execute on real robot

aginika:
system_configure

PR #318, opened 2015-05-01
Add test for picking motion

wkentaro:


wkentaro:
@k-okada this is the test we made.
I merge this.

PR #327, opened 2015-05-01
[verify_object] See stamp to get correct recognition result

wkentaro:


wkentaro:
Issue: https://github.com/start-jsk/2014-semi/issues/311

PR #331, opened 2015-05-01
add left right dividing passthrough

aginika:
Advise from @garaemon

BE CAREFUL THIS WILL CHANGE THE PASSTHROUGH LAUNCH

- divinding to two file.

left_right_dividing

### DIFF
--- a/jsk_2014_picking_challenge/launch/attention_divide.launch
+++ b/jsk_2014_picking_challenge/launch/attention_divide.launch
@@ -1,0 +1,63 @@
+<launch>
+  <arg name="input_cloud" default="/kinect2/depth_highres/points"/>
+  <arg name="input_image" default="/kinect2/rgb/image"/>
+
+  <arg name="custom" default="false"/>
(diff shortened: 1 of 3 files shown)


aginika:
new_passthrough

aginika:
DON'T MERGE THIS UNTIL MOVEMENT BECOME PARALLEL

aginika:
@wkentaro I renamed the filename to attention_divide.launch

wkentaro:
I see.

aginika:
Finally, jsk-ros-pkg/jsk_pcl_ros#891 fix the bug of attention clipper, so we can remove the passthrough filter. Therefore passthrough part and multiplay part in the above diagram has been removed.
It becomes a little bit lighter

k-okada:
i have released both jsk_recognition and jsk_visualization and that would
be available from shadow-fixed very soon, so non pcl developer need to use
source code, use deb packages.

--
◉ Kei Okada

PR #334, opened 2015-05-02
update Baxter self filter params

aginika:
Related #272
Finnaly, the cylinder seems to be enough to remove the pointcloud.
I couldn't find out the reasons.
Kinect2's tf seesm to be a little bit wrong, so please fix it .

both_arm_self_filter


garaemon:
[This patch] (https://github.com/PR2/pr2_navigation/pull/20) support dae in self_filter. You don't need to specify specific cylinder

PR #336, opened 2015-05-02
- [euslisp/model/pod-lowers.l] add :glvertices to compatible with glbody https://github.com/euslisp/jskeus/pull/202

k-okada:


aginika:
@thank you

PR #348, opened 2015-05-03
Change target region as method of *pod*

wkentaro:
Usage
-----
```
(send *baxter* :rarm :inverse-kinematics (make-cascoords :pos (v+ (send *pod* :a) #f(-90 0 0))))
```


wkentaro:
Issue: https://github.com/start-jsk/2014-semi/issues/347

Issue #353, opened 2015-05-03
[error] unbound variable sound_play::soundrequestaction


wkentaro:
In robot-main.l, I had error below::
```
2.irteusgl$ main
[ERROR] [1430642041.454463597]: attempted to getTopic to topic robotsound, which was not previously advertised. call (ros::advertise "robotsound") first.
/opt/ros/indigo/share/euslisp/jskeus/eus/Linux64/bin/irteusgl 0 error: unbound variable sound_play::soundrequestaction in (send #:inst23135 :init "/sound_play" sound_play::soundrequestaction :groupname "speak")

```

aginika:
I couldn't replay on mour.

But I think sound_play 's deb which include aciton_msgs  is not released, so you should clone from github.

k-okada:
if our software depends on unreleased version of sound_play, we need to
update https://github.com/start-jsk/2014-semi/blob/master/README.md, or
better to create custom rosinstall file as
https://github.com/jsk-ros-pkg/jsk_robot/blob/84629a3000f087d58cc2300bd686d1a4a5e804c0/jsk_pepper_robot/pepper.rosinstall

--
◉ Kei Okada

wkentaro:
tnx.

PR #357, opened 2015-05-03
add sound_play install part to README

aginika:
#353 

You need to compile from source.


k-okada:
Please merge after Travis passed, even if you're confident on your commit. Because that's what Travis do for you

aginika:
@k-okada Sorry, I changed only README.md , so I merged...
I will be careful about that from next time

PR #386, opened 2015-05-04
[score_calculator.py] Question each bin scores in work order

wkentaro:


wkentaro:
score_calculator

aginika:
Great!

PR #422, opened 2015-05-07
[robot_main] Don't move for verification unless grabbed

wkentaro:


wkentaro:
Issue: https://github.com/start-jsk/2014-semi/issues/413

PR #466, opened 2015-05-11
add Target Object Interactive Marker Server

aginika:
Maybe for some reasons, dae file doesn't texture the mesh in rviz, so now those are white.
markers


wkentaro:
Can I merge this?
Or should I wait for colored meshes?

aginika:
please wait I need to add other objests too

aginika:
objects

PR #530, opened 2015-05-15
Update baxter white urdf xacro

aginika:


aginika:
Why are these commit ?

PR #563, opened 2015-05-17
[robot-main.l] untabify robot-main.l

wkentaro:


wkentaro:
We should add belows.

.emacs.d/init.el or .emacs
```
 (setq-default indent-tabs-mode nil)
```

.vimrc
```
set expandtab
```

PR #565, opened 2015-05-17
add check state script for before start

aginika:
For  #540
This is still not perfect.
DON'T MERGE UNTIL I DO
check_state


aginika:
littele_update

PR #612, opened 2015-05-18
[robot_main] main do verification enough

wkentaro:


wkentaro:
Issu: https://github.com/start-jsk/2014-semi/issues/604

PR #624, opened 2015-05-19
[pick_object] change range of z to move

wkentaro:


wkentaro:
https://github.com/start-jsk/2014-semi/issues/602

PR #695, opened 2015-05-21
[robot-main.l] Set param after finishing the orders of each arm

wkentaro:


wkentaro:
Issus: https://github.com/start-jsk/2014-semi/issues/689

PR #710, opened 2015-05-21
[robot-main.l] Fix state change after pick_object

wkentaro:


wkentaro:
@k-okada @aginika @iory 
Sorry for the bug in state change in robot_main.
It was very complicated so I simplified it.

PR #740, opened 2015-05-22
Add axis video recorder

wkentaro:
```
rosrun jsk_2014_picking_challenge record_axis_video.py
```


wkentaro:
aaa

k-okada:
Nice, we can move this to upstream package after drc

PR #789, opened 2015-05-24
modified pick-object to avoid objects collision

iory:
modified pick-object to avoid return objects's collision
add pick-objects's argument :n-tried


wkentaro:
See #790

PR #790, opened 2015-05-24
modified pick-object to avoid objects collision

wkentaro:


wkentaro:
Simplifyed diff of https://github.com/start-jsk/2014-semi/pull/789

PR #819, opened 2015-07-16
-  add .rosinstall (jsk_robot)

k-okada:


wkentaro:
the filename .rosinstall causes problem for wstool.
is it possible to change it?

wkentaro:
See #823

PR #823, opened 2015-09-01
Fix test-robot-motion

wkentaro:


wkentaro:
I fixed the robot-motion-test.
Now all motions are tested with one testblock.

wkentaro:
in hydro, the test fails because of missing baxter-utils.l

PR #913, opened 2015-10-13
[jsk_2015_apc_common] gazebo_ros to pass models path to gazebo

wkentaro:
See also: https://github.com/start-jsk/jsk_apc/pull/906


wkentaro:
We need model.sdf and model.config to put the models on gazebo.

PR #918, opened 2015-10-15
[jsk_2015_apc_common] Keep vision timestamp even if transformed

wkentaro:


wkentaro:
no idea why this happen
```
Preparing to unpack .../cmake-data_2.8.12.2-0ubuntu3_all.deb ...
ERROR: cmake-data is broken - called emacs-package-install as a new-style add-on, but has no compat file.
Unpacking cmake-data (2.8.12.2-0ubuntu3) ...
Selecting previously unselected package cmake.
P
```

PR #939, opened 2015-10-19
[jsk_2015_05_baxter_apc] Concatenate head/torso points

wkentaro:


wkentaro:
Maybe merging to point cloud topics is not good for organized point clouds, so I close this pR.

PR #952, opened 2015-10-27
[jsk_2015_05_baxter_apc] Make faster localization in hand

wkentaro:
* use self_filter in bottom


wkentaro:
this achieves

* 6hz points from kinect2
* 3hz segmented image.

PR #955, opened 2015-10-27
[jsk_2015_apc_common] Add model config files to put object models in gazebo world

wkentaro:
gazebo_apc


wkentaro:
for: https://github.com/start-jsk/2015-semi/issues/2

Issue #966, opened 2015-11-17
Test recognition nodes


wkentaro:
- [ ] localization of objects in bin
- [x] localization of object in hand
- [x] recognition of object in hand

wkentaro:
Close for now.

PR #976, opened 2015-11-18
[jsk_2015_05_baxter_apc] Test recognitioin in hand

wkentaro:
Depends on https://github.com/jsk-ros-pkg/jsk_common/pull/1240


wkentaro:
For #966

PR #977, opened 2015-11-18
[jsk_2015_05_baxter_apc] Run test actually

wkentaro:


wkentaro:
need to install jsk_recognition from source.

wkentaro:
Finally fixed.

PR #985, opened 2015-12-05
[jsk_2015_05_baxter_apc] Pass timestamp to recognition method

wkentaro:


wkentaro:
need to wait for jsk_common's release.

PR #1067, opened 2016-02-14
hsv based clustering added

kimhc6028:


knorth55:
I think there are some bugs. see #1049.
you should use inheritance for color_object_matcher. see [here](http://www.geocities.jp/m_hiroi/light/python06.html)

kimhc6028:
https://github.com/start-jsk/jsk_apc/tree/master/jsk_apc2015_common#train-classifier-with-dataset
says directory name of masked image files for berkeley_dataset is berkeley_dataset_mask_applied .
If you copy berkeley_dataset_mask_applied to jsk_2016_01_baxter_apc/data, I think color_histogram_features.py could generate histogram files.
color_object_matcher.py is just copied from jsk_2015. Could you tell me more specifically how to improve this file?

knorth55:
1. In your case, the program is looking in the directories named after each image's name (i.e. masked_data/oreo_mega_stuff/oreo_mega_stuff.jpg)
In my closed PR, the program just search in the data dir (i.e. masked_data/oreo_mega_stuff.jpg)
I'm not sure whether you want to do.

2. Use Class Inheritance (クラス継承）
I think just copy and paste is not wise way. Import color_object_matcher class in jsk_2015_05_baxter_apc and inherit it.
( anyway, you should change jsk_2015 -> jsk_2016 https://github.com/kimhc6028/jsk_apc/blob/hsv_histogram/jsk_2016_01_baxter_apc/node_scripts/color_object_matcher.py#L74 )

kimhc6028:
1. In berkeley dataset, image's name is not the label of the image (i.e. N3_27.jpg, instead of oreo_mega_stuff.jpg). The dataset represents image's label by directory's name it is belong to(i.e. all jpg files under berkeley_dataset_mask_applied/oreo_mega_stuf/ are labeled to oreo_mega_stuf). Thus, I think looking for the directory's name is required
2. I failed to import python files from another directory. Could you teach me how to do it tomorrow?

PR #1069, opened 2016-02-14
object_list modified

kimhc6028:
modified object_list.yml according to
http://rll.berkeley.edu/amazon_picking_challenge/
I am not quite sure which is standard name of objects.


wkentaro:
the names listed in object_list.yaml is standard one. 
some of berkeley's ones are wrong.

PR #1074, opened 2016-02-15
color_histogram_features edited

kimhc6028:
using hsv color histogram


wkentaro:
This breaks the current demo, does not?
I think hsv should be optional otherwise you need to update the trained model file here: https://github.com/start-jsk/jsk_apc/tree/master/jsk_2015_05_baxter_apc/data/histogram_data

PR #1077, opened 2016-02-16
Describe about kinect2 setup versions which works

wkentaro:


wkentaro:
Closes https://github.com/start-jsk/jsk_apc/issues/1016

k-okada:
I have added https://github.com/start-jsk/jsk_apc/blob/master/jsk_2016_01_baxter_apc/rosinstall for kinect2 setup, please choose 
- use wstool and rosinstall 
OR
- remove rosinstall file

wkentaro:
> - use wstool and rosinstall OR
- remove rosinstall file

you mean adding kinect2 setup to https://github.com/start-jsk/jsk_apc/blob/master/jsk_2015_05_baxter_apc/rosinstall ?
But why do I need to remove rosinstall file?

k-okada:
I forget about https://github.com/start-jsk/jsk_apc/blob/master/jsk_2015_05_baxter_apc/rosinstall, do we still need to use v1.0.1 of baxter software? because of real robots?

anywey we'd better to put kinect2 information to one place, 
may be creating kinect.rosinstall at jsk_apc and fix documents will be ok

k-okada:
and remove kinect2 information from rosinstall files.

wkentaro:
> I forget about https://github.com/start-jsk/jsk_apc/blob/master/jsk_2015_05_baxter_apc/rosinstall, do we still need to use v1.0.1 of baxter software? because of real robots?

I specify `v1.1.0` to match it with real robot's one.

> anywey we'd better to put kinect2 information to one place, 
may be creating kinect.rosinstall at jsk_apc and fix documents will be ok
and remove kinect2 information from rosinstall files.

I see. I will do it.

PR #1078, opened 2016-02-16
Add movie of real demo

wkentaro:


k-okada:
basically, whenever you publish video, associate video with them with
published paper, that mean put reference to the paper in the beginning or
the end roll.
That mean you do some experiments, and take video, publish paper, then put
the video on youtube.

--
◉ Kei Okada

wkentaro:
Hmm, what do you mean?
You mean I should not share movie before publishing paper?
What is the difference between it and pictures we usually share on GitHub?

PR #1145, opened 2016-03-14
Use deb on travis test for faster testing

wkentaro:


wkentaro:
27min -> 13min
screen shot 2016-03-14 at 10 30 05 pm

PR #1278, opened 2016-04-12
Upgrade catkin_tools in travis

wkentaro:


wkentaro:
the verbosity is too large.

PR #1325, opened 2016-04-21
[jsk_2016_01_baxter_apc] Softkinetic overlook pose

knorth55:
add softkinetic overlook pose method, the position where the camera can see totally inside of a bin.

softkinetic_overlook_pose

cc. @yuyu2172


knorth55:
@yuyu2172 how about these?

yuyu2172:
I like it.  We can use this code as a template in the future.
After implementing a new segmentation algorithm, we should further investigate arm positions where the algorithm perform well.

BTW, can it see an object near the back of the shelf? From how it looks in the picture, a camera should be a little bit higher to look into objects located near the back. We can look into that in the future.

knorth55:
i tried to avoid occlusion between bin and gripper so that the overlook-pose is a little bit low.
I can make it higher, but unfortunately with the occlusion. 
we should make better position in future

knorth55:
please merge

PR #1331, opened 2016-04-24
[jsk_2016_01_baxter_apc] Add enable_gripper node enabling grippers and managing automatic power off

pazeshun:
From #1297 , #1327 .
This node works with new Arduino firmware @ban-masa will make.


pazeshun:
Now this node works with #1335. Without #1335, this only supports automatic power off.
Please merge this.

PR #1332, opened 2016-04-25
[WIP] add new segmentation algorithm

yuyu2172:
# What I have done

I implemented a segmentation algorithm that is described here. [paper](http://www.robotics.tu-berlin.de/fileadmin/fg170/Publikationen_pdf/Jonschkowski-16-Technical-Report.pdf)

Some part of the algorithm and a public data set are shared [here](https://gitlab.tubit.tu-berlin.de/rbo-lab/rbo-apc-object-segmentation).

I implemented data pre-processing codes and ROS interface which lacked in the public code. 

The algorithm takes RGBD information as input and produce segmented image as output. The first picture illustrates output. The second image illustrates internal data that evaluates probability of pixel assigned to each class. The figures in the second image are rotaed 90 degrees. 

2016_04_24_20_12_bin_g_cheezit_big_original

screenshot from 2016-04-25 10 55 02

# TODOs
+ The images are taken by kinect2 on torso. Run algorithm on data taken from softkinetic camera.
+ Take more examples and analyze performance.

PR #1357, opened 2016-04-28
[jsk_apc2016_common] Pre-process data for segmentation in bin

yuyu2172:
A piece divided from #1332.

This is a set of functions that are used to extract spatial features for segmentation algorithm. More precisely, it extracts two information: 
1. distance between point clouds and the target bin. The distance is the shortest distance between the point clouds and faces of a bounding box that locates at the target. Among six faces of the bounding box, a entrance of the bin is excluded from calculation.  
2. height of the point clouds and the bottom of the target bin.

### results
distance to the target bin
screenshot from 2016-04-28 11 24 19

height
screenshot from 2016-04-28 11 24 50

edit: now includes #1358


knorth55:
+1

PR #1358, opened 2016-04-28
[jsk_apc2016_common] add a script that generates mask image of the target bin

yuyu2172:
Generate region of an image that corresponds to the target bin.


wkentaro:
+1

PR #1368, opened 2016-04-29
[jsk_apc2016_common] add BinInfo.msg and BinInfoArray.msg

yuyu2172:

PR #1370, opened 2016-04-29
[Depends on #1368][jsk_apc2016_common] 	add BinData class that adds extra data to BinInfo msg

yuyu2172:
This class calculates location of corner points which are used to generate a mask image of a bin.
I have decided not to add corner points in Bin.msg's attributes because depending on how you choose to generate the mask image you may not need them.
In other words, BinPython contains codes that depend on algorithms.

Merge after #1368.


knorth55:
I think it's better to rename file. `bin.py` is too general and cannot clearly understand what it is for

yuyu2172:
Thanks for your comment.
How about apc_bin.py?

knorth55:
+1

PR #1377, opened 2016-04-30
[jsk_2016_01_baxter_apc] Test move-arm-to-bin with rosbag for bin boxes

wkentaro:


pazeshun:
To #1378

wkentaro:
@pazeshun Could you please open a issue at [jsk_roseus](https://github.com/jsk-ros-pkg/jsk_roseus/issues) about huge tf erros in roseus when testing with rosbag --loop?

pazeshun:
@wkentaro Ok, I'll open.

pazeshun:
@wkentaro I found the question and answers about that tf errors.
http://answers.ros.org/question/66586/tf_old_data-error-when-i-play-bag-file-for-rgbdslam/
http://answers.ros.org/question/206748/tf_old_data-ignoring-data-from-the-past-for-frame-openni_depth_optical_frame/
I began to think this problem isn't only issue of roseus. I tried `$ rosparam set use_sim_time true`, but got the same tf errors like the second post.
I heard @knorth55 once researched about this issue. @knorth55 Do you remember about this?

PR #1380, opened 2016-04-30
[jsk_2016_01_baxter_apc] check ik solvable x-axis limit 

knorth55:
use apc2016 baxter model
clean up codes


wkentaro:
please add labels and milestone.

wkentaro:
You should run this by catkin run_tests.

knorth55:
this one is for checking region program, so it fails every time.
i made test for baxtereus and plan to add it to this repo with another PR
 https://github.com/jsk-ros-pkg/jsk_robot/pull/566

wkentaro:
So, this is test to be checked by human? Not automated validation?

knorth55:
yes, because we can solve ik  at not every point in bin. we can find out which point we cannot solve ik with this program.

wkentaro:
Ok, could you please output here?

wkentaro:
Ok, could you please upload output here?

knorth55:
ok

knorth55:
checking done
```
:a: solvable x-axis depth is 1207.88
:b: solvable x-axis depth is 1054.34
:c: solvable x-axis depth is 1100.79
:d: solvable x-axis depth is 1373.89
:e: solvable x-axis depth is 1320.34
:f: solvable x-axis depth is 1316.8
:g: there is no ik-unsolvable region
:h: solvable x-axis depth is 1360.0
:i: solvable x-axis depth is 1356.34
:j: solvable x-axis depth is 1350.25
:k: solvable x-axis depth is 1326.64
:l: solvable x-axis depth is 1322.98
```
ik-left
ik-right

wkentaro:
> right gripper looks working bad because i do checking with rotation-axis t.
if we can solve ik like rotation-axis (x but range is 0-90 degree), it would be much better

Could you please add a issue for this?

pazeshun:
@knorth55 
Thank you for great work.
I think the reason why right gripper works bad is right gripper's length.
The direction of the target coordinates of IK in this check is certain, z up, x forward. So the servo on right gripper should rotate. When that servo rotates, right gripper is shorter than left gripper.

knorth55:
@pazeshun 
> Maybe testing two target coordinates(one is z up and x forward, the other is x up and z backward) brings out the performance of the right gripper.
i understood it when i saw the result. i will do it when i have time.

PR #1391, opened 2016-05-02
[jsk_2016_01_baxter_apc] Use ConnectionBasedTransport in RBOSegmentationInBinNode

knorth55:


jsk_2016_01_baxter_apc/node_scripts/rbo_segmentation_in_bin_node.py
```diff
@@ -61,8 +61,8 @@ def _callback(self, cloud, img_msg, camera_info):
         camera2bb_base = self.buffer.lookup_transform(
```
wkentaro: You need to add:
```
def unsubscribe(self):
    self.pc_sub.unregister()
    sefl.cam_info_sub.unregister()
    sefl.img_sub.unregister()
```

knorth55: added

yuyu2172:
+1

PR #1395, opened 2016-05-03
[jsk_apc2016_common] unzoom returned prediction

yuyu2172:
Previously, the image returned by RBO's segmentation algorithm zoomed into mask of a bin.
I fixed it to unzoom from the bin.

#### before
prediction

#### after
prediction

What the returned segment look like with rgb image on background.
color_prediction


pazeshun:
Please add jsk_apc2016_common tag to title

yuyu2172:
Ops. My bad.

knorth55:
+1

PR #1397, opened 2016-05-03
[jsk_2016_01_baxter_apc] add header to rbo topic

knorth55:


yuyu2172:
thanks!

jsk_2016_01_baxter_apc/node_scripts/rbo_segmentation_in_bin_node.py
```diff
@@ -82,8 +82,10 @@ def _callback(self, cloud, img_msg, camera_info):
         # generate a binary image
         self.segmentation()
         try:
-            self.img_pub.publish(self.bridge.cv2_to_imgmsg(
-                    self.predicted_segment, encoding="passthrough"))
+            mask_msg = self.bridge.cv2_to_imgmsg(self.predicted_segment, encoding="passthrough")
+            mask_msg.header.stamp = rospy.Time.now()
```
wkentaro: ?
```
mask_msg.header.stamp = img_msg.header.stamp
```

knorth55: the process took time, so if i do your suggestion, stamp is not correct

wkentaro: > the process took time, so if i do your suggestion, stamp is not correct

What is correct about the stamp for you?
I think the stamp should be when the original input data is generated. As for this node, it is camera image input.

knorth55: i think stamp means when the mask image is generated.
do you know the public definition?

wkentaro: I don't know where the definition is. It is just a common sense in ROS community.
Because the information about when the mask is generated is not useful in most cases.
On the otherhand, the information about when the original input data is generated can be used, for example, in order to apply generated mask to input rgb image by synchronizing the time stamp.
- jsk_pcl_ros: https://github.com/jsk-ros-pkg/jsk_recognition/blob/master/jsk_pcl_ros/src/extract_indices_nodelet.cpp#L104
- pcl_ros: https://github.com/ros-perception/perception_pcl/blob/jade-devel/pcl_ros/src/pcl_ros/filters/filter.cpp#L99
- opencv_apps: https://github.com/ros-perception/vision_opencv/blob/indigo/opencv_apps/src/nodelet/edge_detection_nodelet.cpp#L223

knorth55:
@wkentaro i understand how to use stamp, thank you

wkentaro:
LGTM

PR #1399, opened 2016-05-04
[jsk_2016_01_baxter_apc] Split Transform Listener node

knorth55:
RBO segmentation node split into two, transform listener is separated from RBO node.
Solving#1396


yuyu2172:
@knorth55 
WIP?

knorth55:
closed

PR #1416, opened 2016-05-09
[jsk_2016_01_baxter_apc] fixed codes that use rosparam target bin 

yuyu2172:
two fixes
1. Dealt with the case when rosparam 'target_bin' is indeterminate. 
2. initialization of fromBinInfo no longer depends on target_bin

### DIFF
--- a/jsk_2016_01_baxter_apc/node_scripts/rbo_segmentation_in_bin_node.py
+++ b/jsk_2016_01_baxter_apc/node_scripts/rbo_segmentation_in_bin_node.py
@@ -38,6 +38,8 @@ class RBOSegmentationInBinNode(ConnectionBasedTransport, RBOSegmentationInBin):
         self.sub .unregister()
 
     def _callback(self, sync):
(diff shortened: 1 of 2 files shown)


jsk_2016_01_baxter_apc/node_scripts/rbo_segmentation_in_bin_node.py
```diff
@@ -38,26 +38,30 @@ def unsubscribe(self):
...
+        self.target_object = self.shelf[self.target_bin_name].target
+        self.target_bin = self.shelf[self.target_bin_name]
 
         # mask image
-        self.camera_info = camera_info
+        self.camera_info = sync.cam_info
+        self.camera_model.fromCameraInfo(self.camera_info)
         try:
             img_color = self.bridge.imgmsg_to_cv2(img_msg, "bgr8")
+            img_color = cv2.cvtColor(img_color, cv2.COLOR_BGR2HSV)
```
knorth55: what is this for?

PR #1419, opened 2016-05-09
[jsk_2016_01_baxter_apc] add :check-bin-exist method

knorth55:
check if target bin is exist.
if not skip the target bin.


yuyu2172:
lgtm

PR #1434, opened 2016-05-11
Add document about rules for APC2016

wkentaro:
```
Dear Picking Challenge community,

Attached are three sheets describing the items for the challenge this year. The first is just the item images you received before but with the two toothbrushes shown separately. Next is the item bonus sheet which shows which items there will be two copies of in the competition (2x) and which items are worth bonus points (+1, +2, +3). The final sheet clarifies what the json labels are for each item.

The item bonuses work as follows. Assume that the pick task work order specifies picking the pencil cup from a bin with 4 items (pencil cup + 3 others). Successfully moving the pencil cup to the tote will then be worth:
+ 15 for a pick from a 4 item bin
+3 for pencil cup item bonus
For a total of 18 points for that pick.

We have also attached two python scripts to assist with you with testing and development. The first will generate random pick and stow task input files with the all the items included. The second can test any pick or stow task input file to make sure the contents are correct.

Happy picking,
```

- [x] Add sample images/interface json/output json


knorth55:
+1

wkentaro:
UPdated : https://github.com/wkentaro/jsk_apc/blob/2f2e6d11a6d85501da647e14d0d0ce8cb0fc3223/doc/apc2016_rules.md

knorth55:
thank you very much :)

PR #1435, opened 2016-05-11
Refine milestone document files

wkentaro:
in terms of

- name
- title


pazeshun:
+1
Thank you very much.

PR #1439, opened 2016-05-11
[jsk_apc2016_common] Faster computing for RBO segmentation preprocessing

wkentaro:
## Summary
This is for faster computing in RBO preprocessing stage by:

- Make faster by removing no need loop for spatial features 
- Remove no need assertion and loop in project_points for pythonic & faster
- Cythonize get_spatial_features for faster computing

I also added test.


wkentaro:
@yuyu2172 I'm not sure about how to test the RBO segmentation. (if there is test program, please tell me)
Please review and checks this PR's validity. I believe this PR makes the preprocessing faster.

yuyu2172:
First of all, thanks for writing the code to help us.

There is no test yet. 
The PR is valid and seems to be faster than the previous one.

@knorth55 and I have agreed upon splitting the SIB into multiple nodes, and splitting work is under way #1443.
I think that cpp code written in #1443 can have equal if not more performance gain.

knorth55:
what is status of this PR?

knorth55:
ping

wkentaro:
This is not required now, because all of the heavy operations are done by c++.

PR #1441, opened 2016-05-12
[jsk_apc2016_common] Publish_bin_info now publishes tf frame of bin

yuyu2172:
I did two things in this PR.
Bin's frame did not exist previously. This PR publishes those frames by using BoundingBoxes. The published frame simplifies pre-processing spatial features. Rewrite of the pre-processing is under way. 
 
Also, BinInfo msg now includes header which is beneficial in two ways.
1. frame_id: used to locate frame of the bin
2. stamp: used by message filter.


knorth55:
can you split tf publish function into one another node? I don't see any necessity of having this function in this node.

yuyu2172:
that works with me

knorth55:
what do you mean?

knorth55:
ah, i got it

yuyu2172:
I guess "that works *for* me" should be the correct phrase. I learned something. Thanks.

yuyu2172:
@knorth55 
ping

knorth55:
+1
can you put a screenshot of tf_echo or view_frames?

yuyu2172:
sure

PR #1426, opened 2016-05-11
[jsk_2016_01_baxter_apc] modify func get_mask_img() into one node

knorth55:
rbo_segmentation_node.py have three functions, mask image generation, spatial image generation and segmentation.
our idea is that it is easier to improve and debug if we split the rbo_segmentation_node.py into three node.

PR #1442, opened 2016-05-12
[Depends on #1441][jsk_apc2016_common] add publish_target_bin

yuyu2172:
This node publishes a BinInfo of a target bin.
It is nice to use the published BinInfo because nodes that use it do not need to do condition handling of target_bin_name.


knorth55:
publish_target_bin_info is better, i think

yuyu2172:
right

jsk_apc2016_common/node_scripts/publish_target_bin_info.py
```diff
@@ -0,0 +1,36 @@
...
+            continue
+        bin_info_dict[target_bin_name].header.seq = (
+                bin_info_dict[target_bin_name].header.seq + 1)
+        target_bin_info = bin_info_dict[target_bin_name]
+        target_bin_info.header.stamp = rospy.Time.now()
+        target_bin_info_pub.publish(target_bin_info)
+
+
+if __name__ == '__main__':
+    rospy.init_node('sib_publish_target_bin_info')
```
knorth55: `publish_target_bin_info` is better

yuyu2172: why?

yuyu2172:
a new commit in this PR is 'add publish target_bin'.

knorth55:
i think nodes should be in `jsk_2016_01_baxter_apc` package, not in `jsk_apc2016_common` like #1426

wkentaro:
Maybe it should be installed in `jsk_apc2016_common`. As discussion in
https://github.com/start-jsk/jsk_apc/pull/1332#issuecomment-214097253

knorth55:
i don't understand the difference of these packages.
is these common nodes?

knorth55:
+1

PR #1443, opened 2016-05-12
[jsk_apc2016_common] Spatial Preprocessing Node

yuyu2172:
related #1439

### DIFF
--- a/jsk_apc2016_common/CMakeLists.txt
+++ b/jsk_apc2016_common/CMakeLists.txt
@@ -12,7 +12,13 @@ find_package(catkin REQUIRED COMPONENTS
     jsk_recognition_msgs
     sensor_msgs
     message_filters
+    tf2
+    tf2_ros
+    tf2_eigen
+    pcl_ros
+    cmake_modules
 )
+find_package(Eigen REQUIRED)
 
 ## System dependencies are found with CMake's conventions
 # find_package(Boost REQUIRED COMPONENTS system)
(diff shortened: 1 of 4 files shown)


yuyu2172:
a new commit in this PR is 'add sib_spatial_preprocessing node'. 
It is not the latest commit in this PR due to rebasing. Don't worry about the order of commits.

knorth55:
`sib_spatial_preprocessing.cpp` should be in package `jsk_2016_01_baxter_apc` according to https://github.com/start-jsk/jsk_apc/pull/1332#issuecomment-214097253

and also you should change its name something like `tf_bbox_to_spatial`.
you don't need to specify `sib` or `preprocessing`.

yuyu2172:
@knorth55 
This should be in package jsk_apc2016_common according to #1332 and #1442.

I agree with the name you proposed.

knorth55:
so you think that tf_bbox_to_mask.py might be in jsk_apc2016_common?
i think ot uses kinect  camera position and it depends on robot.
and in that case, it should be in jsk_2016_01_baxter_apc package

yuyu2172:
Yes, I think that tf_bbox_to_mask should be moved to jsk_apc2016_common.

I disagree your definition of `robot specific code`.
The code, in principle, works on any robots with tf and a camera, such as PR2.
That is the reason why the code should `not`

yuyu2172:
@wkentaro 
ping

PR #1448, opened 2016-05-12
[jsk_2016_01_baxter_apc] Don't overload gripper servo when placing object

pazeshun:
When placing an object in order bin, the gripper servo is always powered on. But if the object is very heavy, gripper servo is overloaded when gripper direction and gravity direction aren't the same while bringing the object.
This PR will fix this using `:gripper-servo-off` in `:move-arm-body->order-bin` and `:gripper-servo-on` in `:spin-off-by-wrist`


pazeshun:
The servo was broken because of this overload.
I replace it with new one.

ban-masa:
+1

PR #1453, opened 2016-05-12
[jsk_2016_01_baxter_apc] fix bug in main.l and add euslint test

knorth55:


wkentaro:
Please try https://github.com/start-jsk/jsk_apc/tree/euslint
and add test with `jsk_tools_add_shell_test` in CMakeLists.txt.

knorth55:
like this?
i tried euslint with error code and it doesn't pass error result to catkin.

wkentaro:
lgtm

knorth55:
it seems catkin does not trap euslint error, doesn't it?
I tried without fix-bug commit , andit passed run_test in my env.

pazeshun:
Does this bug cause something bad? If it does, I'll cherry-pick this PR to gripper-v2 branch immediately.
If it doesn't, I think I don't have to hurry.

wkentaro:
> Does this bug cause something bad? If it does, I'll cherry-pick this PR to gripper-v2 branch immediately.
If it doesn't, I think I don't have to hurry.

@knorth55

Issue #1456, opened 2016-05-12
[jsk_apc_2016_baxter_apc] robot self filter does not work properly


knorth55:
Because `robot_description` is modified, `robot_self_filter` does not work properly so that it failed to verify object in hand.
Upper: before self filtered
Lower: after self filtered 
![before_filter](https://cloud.githubusercontent.com/assets/9300063/15222662/cb140d62-18ab-11e6-8be7-06bf9675054c.png)
![after_filter](https://cloud.githubusercontent.com/assets/9300063/15222680/dec6847a-18ab-11e6-899c-dbc016240dde.png)

knorth55:
@wkentaro do you know how to visualize self filtered region?

knorth55:
robot self filter works properly with stl model?

knorth55:
i convert stl models to collada models, but it does not work

knorth55:
@pazeshun can you cherry-pick the commit and try it on gripper-v2 branch?

pazeshun:
@knorth55
ok, I'll try it for Saturday demo

wkentaro:
This seems to be introduced by new gripper model, so I change the assignee to @pazeshun .

pazeshun:
I made a new branch for debugging from gripper-v2 branch and cherry-picked a part of #1458 
https://github.com/pazeshun/jsk_apc/tree/fix-self-filter-2016
In debugging, I realized that robot_self_filter works when I comment out `right_gripper_base` and `right_gripper_vacuum_pad_base` in self_filter.yaml.
![screenshot from 2016-05-14 11 31 20](https://cloud.githubusercontent.com/assets/14994939/15265768/068b21be-19c9-11e6-9add-fb20bffb9ded.png)

pazeshun:
I changed collision meshes of `right_gripper_base` and `right_gripper_vacuum_pad_base` from STL data to URDF boxes.
Then robot_self_filter worked.
![screenshot from 2016-05-14 17 28 28](https://cloud.githubusercontent.com/assets/14994939/15267358/de49df42-19fa-11e6-9c42-a909cef92567.png)
I'll make suitable collision meshes of these two links using URDF boxes and cylinders.

pazeshun:
Also, I realized that padding of grippers in self_filter.yaml of 2015 is too big for new right gripper.
[...]

PR #1458, opened 2016-05-12
[jsk_2016_01_baxter_apc] add robot self filter for apc2016 robot model

knorth55:
#1456


pazeshun:
+1
I'll debug why self filter didn't move with right arm.

Issue #1457, opened 2016-05-12
Demo on 2016-05-13/14: Segmentation in bin (version 0.3)


knorth55:
****

- @yuyu2172 @knorth55 
- Time: 14:00 - 14:15

(update by @wkentaro)
*****

work on this branch.
https://github.com/knorth55/jsk_apc/tree/sib-demo

knorth55:
this is temporary branch for demo on 05-13 and  05-14.
it works only with left arm and use segmentation algorithm with kinect2_torso. 

```bash
$ roslaunch jsk_2016_01_baxter_apc baxter.launch
$ roslaunch jsk_2016_01_baxter_apc setup_torso.launch
$ roslaunch jsk_2016_01_baxter_apc segmentation_in_bin.launch json:=$(rospack find jsk_2015_05_baxter_apc)/json/layout_12.json
$ roslaunch jsk_2016_01_baxter_apc main.launch json:=$(rospack find jsk_2015_05_baxter_apc)/json/layout_12.json
```

wkentaro:
duplicate of #1452 ?

wkentaro:
You want to split 0.3 from0.4?

knorth55:
yes

yuyu2172:
finished #1454 #1464.

yuyu2172:
@knorth55 

I rebased knorth/sib-demo and pushed a branch called sib-may-fes to origin.
The branch is ready to run with following code.

```
$ roslaunch jsk_2016_01_baxter_apc baxter.launch
$ roslaunch jsk_2016_01_baxter_apc setup_torso.launch
$ roslaunch jsk_2016_01_baxter_apc segmentation_in_bin.launch json:=$(rospack find jsk_2015_05_baxter_apc)/json/layout_12.json
```

wkentaro:
@knorth55 Is that possible to pick object based on the recognition result of RBO segmentation?

yuyu2172:
@knorth55 
I am not thinking of merging this branch to master in the future.
What do you think?

knorth55:
on `knorth55/sib-demo`, we can pick with rbo segmentation.
i think it's ok to merge this to master branch.

yuyu2172:
@wkentaro 
We are going to merge `sib-may-fes` to master.

knorth55:
@yuyu2172 thank you so much

PR #1465, opened 2016-05-13
[jsk_apc2016_common] cloud_to_spatial_features deal with the case when tf of bin does not exist

yuyu2172:
bin tf may not exist at the point when this node is initiated.


wkentaro:
Why not show warning message?

yuyu2172:
that makes sense.

yuyu2172:
fixed

yuyu2172:
#1457 
integrated to `sib-may-fes`

PR #1466, opened 2016-05-13
[jsk_apc2016_common] fix tf_bbox_to_mask slow down

yuyu2172:
The node publishes mask_msg with old time stamp.
For the time being, I fixed time stamp of the published message.

I think this should be fixed as soon as demo ends.


yuyu2172:
This patch works *fine* with kinect2, but fails to work on softkinetic.

The problem occurs when the time on a time stamp of a published message does not reflect actual point of the time when tf of camera is looked up. 
Since tf of kinect2 does not change temporarily, the problem does not manifest on Kinect2. I am pretty sure that on softkinect, it is otherwise.

yuyu2172:
resolved. thanks kentaro

yuyu2172:
#1457 
integrate to `sib-may-fes`

PR #1467, opened 2016-05-13
[jsk_apc2016_common] Topic synchronizer for RBO segmentation in bin node

yuyu2172:
rbo_segmentation_in_bin node takes 4 images. When message filter is used, it crashes probably due to high workload.

I added synchronizing node written in cpp that publishes a message with 4 images so that python node does not need to use message filter.

Issue #1454, opened 2016-05-12
Visualization of segmentation in bin


yuyu2172:
Currently, there is no tool that visualizes segmentation in bin algorithm.
Visualization is necessary in two ways.
1. debugging 
2. demonstration on the Saturday.

I am considering of adding two visualizations.
1. probabilistic heat map of segmentation (how likely each pixel should be assigned to objects)
2. final outcomes (position and pose of the center of the target object)

wkentaro:
maybe You dont need to create new tools, but just publishes it. 
check nodes in jsk_perception for images, and rviz plugins. 
or ask me.

wkentaro:
Possibly below nodes are useful:

> 1. probabilistic heat map of segmentation (how likely each pixel should be assigned to objects)

- [colorize_float_image](http://jsk-docs.readthedocs.io/en/latest/jsk_recognition/doc/jsk_perception/nodes/colorize_float_image.html)

> 2. final outcomes (position and pose of the center of the target object)

- [rviz pose_array plugin](http://wiki.ros.org/rviz/DisplayTypes/PoseArray)
- [image_cluster_indices_decomposer](http://jsk-docs.readthedocs.io/en/latest/jsk_recognition/doc/jsk_perception/nodes/image_cluster_indices_decomposer.html)
- [label_image_decomposer](http://jsk-docs.readthedocs.io/en/latest/jsk_recognition/doc/jsk_perception/nodes/label_image_decomposer.html)
- [draw_rect_array](http://jsk-docs.readthedocs.io/en/latest/jsk_recognition/doc/jsk_perception/nodes/draw_rect_array.html)
- [tile_image](http://jsk-docs.readthedocs.io/en/latest/jsk_recognition/doc/jsk_perception/nodes/tile_image.html)

yuyu2172:
Thank you!

yuyu2172:
done

PR #1468, opened 2016-05-13
[Depends on #1467][jsk_2016_01_baxter_apc] add a launch file that visualizes sib

yuyu2172:
#1454


wkentaro:
Please add image how it looks like.

yuyu2172:
#1457 
integrate to `sib-may-fes`

PR #1487, opened 2016-05-17
[jsk_2016_01_baxter_apc] Update softkinetic camera calibration files

knorth55:


knorth55:
with this commit, `softkinetic/rgb/camera_info` provides correct resolution

PR #1494, opened 2016-05-17
Revert "[jsk_apc2016_common] publish_bin_tf now uses tf2_ros static_tf_publisher"

wkentaro:
Reverts start-jsk/jsk_apc#1493


wkentaro:
requested by @yuyu2172

k-okada:
@yuyu2172 what was the problem?

--
◉ Kei Okada

yuyu2172:
There was a bug.
It only instantiates one publisher even though it should instantiate twelve.

k-okada:
how can I reproduce this "bug"

--
◉ Kei Okada

PR #1498, opened 2016-05-18
[jsk_2016_01_baxter_apc] Fix return_object for right arm

### LINKED ISSUE #1489: [jsk_2016_01_baxter_apc] Collision between bin top and right gripper in return_object
I thought this is fixed in #1478 , but occurred in Bin f.

pazeshun:
Fixes #1489 
- Add two more midpoints to pass
- Use mapcar to improve extendability


ban-masa:
:+1:

PR #1512, opened 2016-05-20
[jsk_2016_01_baxter_apc] Don't send angle-vector if IK fails in approaching and lifting

### LINKED ISSUE #1479: [jsk_2016_01_baxter_apc] IK fails due to #1451
Bin cなどの位置の高いBinで、 #1451 で導入した動きをBinの奥の方にある物品に対してした際、IKが失敗することがある。 #1451 で導入した動きが、アプローチできる物品の範囲を狭めているのではないか。

pazeshun:
From #1470 
Fixes #1479 
If IK fails when robot tries to approach objects straight down or lift objects, give up to do these motions.


ban-masa:
+1

Issue #1519, opened 2016-05-23
Make Segmentation In Bin Failproof


yuyu2172:
When SIB fails to produce valid target mask (i.e. publishes an image with all zero), the predicted position is dangerous. For example, in the image below, SIB failed and the target position is outside of the target bin.

![screenshot from 2016-05-23 20 39 28](https://cloud.githubusercontent.com/assets/2062128/15469665/bc709cc0-2126-11e6-91fd-e32ea47f134d.png)

wkentaro:
This should be fixed before the demo on 2016-05-27. Please work on this @yuyu2172 .
I think it does not take time.

wkentaro:
Is this done?

yuyu2172:
done.
Sorry for late reply.

PR #1518, opened 2016-05-23
[Depends on #1515] [jsk_2016_01_baxter_apc] Avoid bug in robot-interface

pazeshun:
The last commit of this PR fixes #1517 
When jsk-ros-pkg/jsk_pr2eus#227 is fixed, that commit is no more needed.
The other commits are the same as #1515

PR #1523, opened 2016-05-23
[jsk_apc2016_common] rbo_segmentation_in_bin_node returns nothing when it fails to predict anything

yuyu2172:
#1519 

@knorth55, @pazeshun 
We need to change `main.l` to handle cases when rbo_sib fails.

### DIFF
--- a/jsk_apc2016_common/node_scripts/rbo_segmentation_in_bin_node.py
+++ b/jsk_apc2016_common/node_scripts/rbo_segmentation_in_bin_node.py
@@ -65,6 +65,8 @@ class RBOSegmentationInBinNode(ConnectionBasedTransport):
         except CvBridgeError as e:
             rospy.logerr('{}'.format(e))
 
+        self.exist3d_img = (self.dist_img != 0)
+
         target_bin_name = rospy.get_param('~target_bin_name')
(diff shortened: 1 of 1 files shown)


pazeshun:
I think we should change main.l based on #1518

knorth55:
now, if it returns nothing, `main.l` do not try to pick. 
https://github.com/start-jsk/jsk_apc/blob/master/jsk_2016_01_baxter_apc/euslisp/jsk_2016_01_baxter_apc/baxter-interface.l#L334-L338

pazeshun:
In that case, we should use old segmentation system in stead of giving up, I think.

knorth55:
i think it depends on how SIB node's msg looks like.

when it cannot get result and it doesnot send a msg, there is no problem.
However, it cannot get result but send a msg like NaN, it cause a problem.
we can stop this error in SIB node.

yuyu2172:
`:recognize-objects-segmentation-in-bin` in `baxter-interface.l` keeps running in while loop when SIB returns nothing.
This method should timeout when nothing is returned and proceed.

yuyu2172:
@knorth55 
ping

knorth55:
pong

PR #1525, opened 2016-05-24
[jsk_2016_01_baxter_apc] add proper depends on baxter.urdf in CMakeLists

knorth55:
with this PR, all xacro's change is deteceted, and baxter.urdf is updated when `catkin build` is executed.


pazeshun:
+1

Issue #1528, opened 2016-05-24
[jsk_2016_01_baxter_apc] Build error due to long file name


pazeshun:
When I building jsk_2016_01_baxter_apc, I get errors like following
```
CMake Error: Cannot open file for write: /home/pazeshun/apc2016/apc_ws/build/jsk_2016_01_baxter_apc/CMakeFiles/_run_tests_jsk_2016_01_baxter_apc_shell-test__home_pazeshun_apc2016_apc_ws_src_start-jsk_jsk_apc_jsk_2016_01_baxter_apc_scripts_euslint__home_pazeshun_apc2016_apc_ws_src_start-jsk_jsk_apc_jsk_2016_01_baxter_apc_euslisp_examples_picking-with-clustering.l.dir/build.make.tmp
CMake Error: : System Error: File name too long
CMake Error: Could not create /home/pazeshun/apc2016/apc_ws/build/jsk_2016_01_baxter_apc/CMakeFiles/_run_tests_jsk_2016_01_baxter_apc_shell-test__home_pazeshun_apc2016_apc_ws_src_start-jsk_jsk_apc_jsk_2016_01_baxter_apc_scripts_euslint__home_pazeshun_apc2016_apc_ws_src_start-jsk_jsk_apc_jsk_2016_01_baxter_apc_euslisp_examples_picking-with-clustering.l.dir/cmake_clean.cmake
```
It is due to long file name.
When I see default limitation of file name length,
```
$ getconf NAME_MAX .
255
$ getconf PATH_MAX .
4096
```
~~I think these have no problems.~~

pazeshun:
I think `_run_tests_jsk_2016_01_baxter_apc_shell-test__home_pazeshun_apc2016_apc_ws_src_start-jsk_jsk_apc_jsk_2016_01_baxter_apc_scripts_euslint__home_pazeshun_apc2016_apc_ws_src_start-jsk_jsk_apc_jsk_2016_01_baxter_apc_euslisp_examples_picking-with-clustering.l.dir` exceeds `NAME_MAX`

pazeshun:
NAME_MAX is fixed by system, so we can't fix this problem.
I'll move work space to upper directory.

pazeshun:
Fixed.

k-okada:
> NAME_MAX is fixed by system, so we can't fix this problem.

how did you 'fixed by system'

pazeshun:
Fixed by Linux kernel, isn't it?
[...]

PR #1531, opened 2016-05-25
[jsk_2016_01_baxter_apc] euslint target name shorten

knorth55:
Solve #1528


wkentaro:
Why not u use rosrun for euslint?

wkentaro:
I will fix it.

PR #1536, opened 2016-05-25
[jsk_2016_01_baxter_apc] work_order modified to apply for apc2016 objects

knorth55:
MERGE after #1546

Issue #1354, opened 2016-04-27
Graspability Definition for APC2016 objects


pazeshun:
Like #1225 
It isn't in my milestone for now, but what I should do in the near future.

knorth55:
Define Graspability : Level 1-4

Graspability|Description
------|------
**4**|Unable to grasp 
**3**|Difficult to grasp <br> Need complex grasp planning
**2**|A little Difficult to grasp <br> Need idea to grasp but not complex
**1**|Easy to grasp

pazeshun:
@knorth55 Thank you

wkentaro:
This is necessary for version 1.0. Please work on this @pazeshun .

pazeshun:
@wkentaro I'll do this in this week

PR #1546, opened 2016-05-26
[jsk_apc2016_common] object_data_2016 yaml style fixed

knorth55:


wkentaro:
Please update the test.

wkentaro:
And include this kind of PR to #1536 next time.

wkentaro:
This is already included in #1536. duplicated

knorth55:
i think it is better to check separately, so i separated into two PR

knorth55:
#1536 and this PR solve different issue, it should be different PR

wkentaro:
> #1536 and this PR solve different issue, it should be different PR

I don't think they are solving same issue. I'm not sure why it should be separated..
"Use graspability after https://github.com/start-jsk/jsk_apc/issues/1354"

wkentaro:
> Please update the test.

knorth55:
> I don't think they are solving same issue. I'm not sure why it should be separated..


there are two issues
- modify work_order to apply for apc2016 #1536
- object_data yaml style is not proper this PR

knorth55:
> "Use graspability after #1354"

is this a comment for #1536 ?

knorth55:
i will ipdate the test

knorth55:
updated

pazeshun:
I want to define graspability for new gripper based on this PR, so I prefer separating.

knorth55:
@wkentaro ping

Issue #1437, opened 2016-05-11
Use single nodelet manager for each cameras


wkentaro:
## Summary
Currently it is separated, and this makes point cloud communication slow.

- kinect2_torso.launch: ``kinect2_torso``
- setup_torso.launch: ``doura_manager``
- setup_head.launch: ``kinect2_head``

## Possible Solutions

- Use standard nodelet in [kinect2_bridge.launch](https://github.com/wkentaro/iai_kinect2/blob/jsk_apc/kinect2_bridge/launch/kinect2_bridge.launch#L27).
- Use [`jsk_topic_tools/standalone_complexed_nodelet`](http://jsk-docs.readthedocs.io/en/latest/jsk_common/doc/jsk_topic_tools/lib/standalone_complexed_nodelet.html#apis) in all launch files in `jsk_2016_01_baxter_apc`.

wkentaro:
this is done.

PR #1553, opened 2016-05-26
[jsk_2016_01_baxter_apc] Use standalone_complexed_nodelet for setup_softkinetic.launch

wkentaro:
this is necessary to fix bond errors.


knorth55:
#1437

PR #1558, opened 2016-05-26
[jsk_2016_01_baxter_apc] remove duplicated rbo_segmentation_in_bin_node.py

knorth55:


pazeshun:
I think duplicate label means duplication of PR or commits.
If I'm wrong, then don't worry about this comment.

knorth55:
i thought so, too. 
but i picked it because it is same term.

wkentaro:
**duplicate** tag is for duplicated issues or pull requests.

PR #1559, opened 2016-05-26
[jsk_2016_01_baxter_apc] detect :inverse-kinematics nil return

knorth55:
critical error occurs. main.l process has died.


pazeshun:
+1
Please tell me about the condition in which IK failed like #1470

knorth55:
bin k

pazeshun:
I want to know where is target coordinate
We have to do experiment to know that

knorth55:
condition is same as #1650

PR #1588, opened 2016-05-30
fix failing remove gripper link from link-list, (member 'string' list…

k-okada:
…) requries (member 'string' list #'equal)

PR #1589, opened 2016-05-30
Fix failing to remove gripper link from link-list

### LINKED ISSUE #1586: Gripper joint moves when solving IK
From the movie of #1560 
@k-okada 
I realize that the gripper joint moves little when solving IK after #1529 . The screenshot below shows the movement of gripper joint when solving IK by `(send *ri* :ik->bin-entrance :rarm :k :gripper-angle 0)`
screenshot from 2016-05-29 17 43 48
Why does this happen?
I think this will cause unexpected movements and they are sometimes dangerous.
If we can't fix this, we should revert #1529 and #1551

pazeshun:
Closes #1586


pazeshun:
From #1588 
Remove tabs included in `tesk-ik.l` in #1588

PR #1600, opened 2016-05-31
add timeout in method :recognize-objects-segmentation-in-bin

### LINKED ISSUE #1587: If rbo_segmentation_in_bin_node returns nothing, the robot motion stops
Thanks to #1523 , we can prevent unexpected movement of the robot.
But if rbo_segmentation_in_bin_node returns nothing, the robot motion may stop.
The reason of this is on https://github.com/start-jsk/jsk_apc/blob/master/jsk_2016_01_baxter_apc/euslisp/jsk_2016_01_baxter_apc/baxter-interface.l#L244-270 which shows `:recognize-objects-segmentation-in-bin` doesn't finish until messages come from both of `box-topic` and `com-topic`.

I think we have two solutions:
1. Fix `baxter-interface.l` to use `one-shot-subscribe` and timeout in `:recognize-objects-segmentation-in-bin`
2. Fix `rbo_segmentation_in_bin_node.py`
I think the latter solution is better because `:recognize-objects-in-bin` uses the same program as current `:recognize-objects-segmentation-in-bin` to get messages. I don't know much about `:recognize-objects-in-bin` and topics subscribed by it, but `:recognize-objects-in-bin` returns even though no object is found.
@knorth55 @yuyu2172 What do you think?

knorth55:
Closes #1587

### DIFF
--- a/jsk_2016_01_baxter_apc/euslisp/jsk_2016_01_baxter_apc/baxter-interface.l
+++ b/jsk_2016_01_baxter_apc/euslisp/jsk_2016_01_baxter_apc/baxter-interface.l
@@ -242,7 +242,7 @@
       (sethash bin _objects-in-bin-boxes (send box-msg :boxes))
       (sethash bin _objects-in-bin-coms com-msg)))
   (:recognize-objects-segmentation-in-bin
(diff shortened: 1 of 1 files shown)


pazeshun:
+1

PR #1604, opened 2016-06-01
publish_bin_info updates when json is updated

yuyu2172:
publish_bin_info publishes bin-info thta is specified by the current json rosparam


knorth55:
:+1:

PR #1611, opened 2016-06-01
Bin shelf model

yuyu2172:
I measured kiva pod in the real world, and updated model represented by bounding boxes.

In recognition tasks, bounding box should model as real as possible. However, in grasping tasks, that is not the case as bounding boxes are used for collision detection which requires a more conservative model.
Therefore, I decided to make two .yaml files that model kiva pod where each has a different purpose.


k-okada:
we had trouble on travis as described in https://github.com/start-jsk/jsk_apc/pull/1610#issuecomment-223026149, please continue working on your local branch , instaed of waiting for mergeing your PR

yuyu2172:
Okada sensei. Thank you for you comment!

knorth55:
:+1:

wkentaro:
> In recognition tasks, bounding box should model as real as possible. However, in grasping tasks, that is not the case as bounding boxes are used for collision detection which requires a more conservative model.
Therefore, I decided to make two .yaml files that model kiva pod where each has a different purpose.

I'm not sure why you need two different yaml file.
Could you please clarify this? Why the accurate shelf model for recognition does not work in motion planning in euslisp?

yuyu2172:
stopeed using two yaml

pazeshun:
I found that this model works for arm camera, but doesn't work for `:ik->bin-entrance` and so on.
The center point of bin entrance is completely different from real shelf, at least for arms.

knorth55:
> I found that this model works for arm camera

do you mean it works properly with `:move-arm-body->bin-overlook-pose`?

pazeshun:
Oh, I didn't test it.
For now, I tested Bin h only with `picking-with-clustering.l` in #1633 .
I'll test other bin and `picking-with-sib.l`

jsk_apc2016_common/config/bin_lower_shelf.yaml
```diff
@@ -1,7 +1,7 @@
-initial_pos_list: [[-0.27, 0.25, -0.41], [-0.24, -0.01, -0.41], [-0.24, -0.32, -0.41], [-0.27, 0.25, -0.12], [-0.24, -0.01, -0.12], [-0.24, -0.32, -0.12]]
+initial_pos_list: [[-0.22, 0.280, -0.116], [-0.22, 0, -0.116], [-0.22, -0.280, -0.116], [-0.22, 0.280, -0.361], [-0.22, 0, -0.361], [-0.22, -0.280, -0.361]]
 initial_rot_list: [[0,0,0], [0,0,0], [0,0,0], [0,0,0], [0,0,0], [0,0,0]]
 initial_quat_list: [[0,0,0,1], [0,0,0,1], [0,0,0,1], [0,0,0,1], [0,0,0,1], [0,0,0,1]]
-dimensions: [[0.38, 0.23, 0.20],[0.38, 0.27, 0.20],[0.38, 0.23, 0.20], [0.38, 0.23, 0.18],[0.38, 0.27, 0.18],[0.38, 0.23, 0.18]]
+dimensions: [[0.37, 0.248, 0.192],[0.37, 0.306, 0.192],[0.37, 0.248, 0.192], [0.37, 0.248, 0.218],[0.37, 0.306, 0.218],[0.37, 0.248, 0.218]]
 frame_id_list: [kiva_pod_base, kiva_pod_base, kiva_pod_base, kiva_pod_base, kiva_pod_base, kiva_pod_base]
-prefixes: [bin_j, bin_k, bin_l, bin_g, bin_h, bin_i]
+prefixes: [bin_g, bin_h, bin_i, bin_j, bin_k, bin_l]
```
pazeshun: Don't change the order of prefixes.
If you do this, the correspondence of bin to bin name is changed.

pazeshun:
I wrote the cause of `:ik->bin-entrance` failing in a line note.
Bin k and Bin h were swapped.

pazeshun:
**What happened to me**
When I commanded the robot to approach Bin h using `picking-with-clustering.l`, the right gripper entered Bin k.
After debugging, I found that `(send *ri* :ik->bin-entrance :rarm :h)` brought the gripper to Bin k. Like this, Bin g and j, Bin i and l were swapped.
For now, I don't know why.

pazeshun:
Maybe I found the reason.

https://github.com/start-jsk/jsk_apc/blob/master/jsk_2016_01_baxter_apc/euslisp/jsk_2016_01_baxter_apc/baxter-interface.l#L196 presupposes that the order of `bin_gtol_clipper/output/box_array` is Bin j, Bin k, Bin l, Bin g, Bin h and Bin i.
But in this PR, that order is changed to Bin g, Bin h, Bin i, Bin j, Bin k and Bin l.

@wkentaro How did you decide that order?
@yuyu2172 Do you want to use new order?

yuyu2172:
@pazeshun 
Prefixes of the new order is consistent with ``upper_bin_shelf.yaml``. The old order can become a cause of error, and I wanted to change it.

pazeshun:
@yuyu2172 
Thank you for reply.
For the time being, I'll make new PR including this PR.
@wkentaro Do you think new order will cause other errors?

wkentaro:
> @wkentaro How did you decide that order?

That is defined at here: https://github.com/start-jsk/jsk_apc/blob/master/jsk_apc2015_common/config/bin_lower_shelf.yaml#L6

wkentaro:
> @wkentaro Do you think new order will cause other errors?

I don't think so, but I cannot be responsible for it.

pazeshun:
@wkentaro Thank you.
I think @yuyu2172 's suggestion is reasonable, so I also want to change the order.
I'll check `bin_gtol_clipper/output/box_array`'s subscribers.

pazeshun:
I made #1638 to fix problems.
@yuyu2172 Is it ok to move?

yuyu2172:
@pazeshun 
Do you mean we close this PR?

pazeshun:
@yuyu2172 Yes, if you agree.

PR #1610, opened 2016-06-01
use travis not jenkins

knorth55:

PR #1612, opened 2016-06-01
fix travis/jenkins problem

k-okada:
- fix #1610 : disable test to check if download_test_data is a problem
- use travis not jenkins


pazeshun:
Can we close this?

k-okada:
Closed

PR #1626, opened 2016-06-03
Fix gripper joint trajectory action server

pazeshun:
This PR will make following changes:
- Gripper joint trajectory action server will fill time_from_start in feedback published by it
- Gripper joint trajectory action server will publish feedback continuously even if now isn't in command points
- Gripper joint trajectory action server will sleep until trajectory start time when it receive trajectory

Thanks to them, #1624 will be fixed.


ban-masa:
+1

PR #1630, opened 2016-06-03
publish_bin_info publishes messages with headers

yuyu2172:


yuyu2172:
@wkentaro 
please merge

PR #1633, opened 2016-06-06
Enable to use clustering instead of SIB

pazeshun:
Temporarily we need this to run `picking-with-clustering.l`.
I want to use `picking-with-clustering.l` to test grippers easily.


pazeshun:
I confirmed `picking-with-clustering.l` works. I will test `picking-with-clustering.l`

pazeshun:
I confirmed `picking-with-sib.l` works. Please review this.

knorth55:
:+1:

PR #1654, opened 2016-06-09
[WIP] Updates of arduino firmware for new gripper

ban-masa:
This PR should be tested with new grippers before merge.

### DIFF
--- a/jsk_2016_01_baxter_apc/firmware/gripper_front_manager_right.ino
+++ b/jsk_2016_01_baxter_apc/firmware/gripper_front_manager_right.ino
@@ -50,7 +50,7 @@ inline float rad_to_deg(float rad)
 
 void set_angle(void)
 {
-    myservo.write((int)(-rad_to_deg(servo_angle_msg.data) + 90.0));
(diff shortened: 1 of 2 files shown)


jsk_2016_01_baxter_apc/firmware/gripper_front_manager_right.ino
```diff
@@ -50,7 +50,7 @@ inline float rad_to_deg(float rad)
 
 void set_angle(void)
 {
-    myservo.write((int)(-rad_to_deg(servo_angle_msg.data) + 90.0));
+    myservo.write((int)(rad_to_deg(servo_angle_msg.data) - 90.0));
```
pazeshun: Is this for gripper-v3?
I think this shouldn't be fixed until right gripper becomes gripper-v3

ban-masa:
Yeah, this PR is for gripper-v3.
I thought it wasn't a big problem because firmware of right arduino was already written.
Should I make PR for each arm separately?
In that case, I'll separate this PR.

pazeshun:
I think separating is better.

ban-masa:
OK.
Close this PR

PR #1659, opened 2016-06-09
[WIP] add offset to z direction of bbox

yuyu2172:


jsk_apc2016_common/include/sib/cloud_to_spatial_features.h
```diff
@@ -18,6 +18,8 @@
 #include <pcl_ros/point_cloud.h>
 #include <pcl_ros/transforms.h>
 
+#define Z_OFFSET 0.01
```
pazeshun: Use `const` global variable instead of `#define`

knorth55:
👍

jsk_apc2016_common/include/sib/cloud_to_spatial_features.h
yuyu2172: fixed

yuyu2172:
tuning params

pazeshun:
+1 but we should know why this offset is needed.

yuyu2172:
need_for_offeset

yuyu2172:
After adjusting kinect and kiva_pod_base, this PR became unnecessary.

PR #1655, opened 2016-06-09
Make left gripper interface

pazeshun:
~~Works with #1654~~
Works with #1667

PR #1656, opened 2016-06-09
[Depends on #1662] Add left gripper-v3 to recognition flow

pazeshun:

PR #1662, opened 2016-06-09
[Depends on #1655] Add gripper-v3 model

pazeshun:
Make URDF model including gripper-v3 on left arm:
screenshot from 2016-06-10 01 21 29
Visual model of gripper-v3:
screenshot from 2016-06-10 01 22 30
Collision model of gripper-v3:
screenshot from 2016-06-10 01 23 58


pazeshun:
Also, make Eus robot model including gripper-v3:
screenshot from 2016-06-10 01 29 13
We can rotate gripper properly on simulation:
screenshot from 2016-06-10 01 29 47
end-coords of larm is proper:
screenshot from 2016-06-10 01 56 06

pazeshun:
Without #1655 , this won't work in real robot.

ban-masa:
I think there is no mistake, but I'm not sure about xacro...
+1

pazeshun:
Tested in real robot with #1655 , #1656, #1663. No more WIP.
Please merge this.

PR #1679, opened 2016-06-11
Fix encoding of in bin mask: 8UC1 -> mono8

wkentaro:
- 8uc1 cannot be seen with image_view
- jsk_perception/apply_mask does not work with 8uc1


knorth55:
👍

PR #1680, opened 2016-06-12
use bin_info_publish node for :recognize-bin-boxes

knorth55:
with this PR, we do not need to use kinect_head


knorth55:
please merge this

PR #1709, opened 2016-06-15
split publish_bin_bbox from publish_bin_info and rename to clear name

knorth55:


jsk_apc2016_common/node_scripts/publish_bin_info.py
```diff
@@ -21,47 +21,35 @@ def __init__(self):
...
+        rospy.Timer(duration, self.bin_info_publlish)
+        rospy.spin()
+
+    def bin_info_publlish(self, event):
+        json = rospy.get_param('~json', None)
+
+        # update bin_info_arr only when rosparam: json is changd
+        if self.json_file != json:
+            if not os.path.isfile(json) or json[-4:] != 'json':
+                rospy.logwarn('wrong json file name')
```
yuyu2172: You need to opt out from the method when json is invalid.

knorth55: fixed

PR #1712, opened 2016-06-16
Node to collect sib data

yuyu2172:


yuyu2172:
I will add functionality to collect normals in the future after I add a normal publisher in sib pipeline.

PR #1713, opened 2016-06-16
labelme specialized for APC

yuyu2172:
# Features
1. After finish labeling, it automatically opens labelme with the next data in stack
2. save labeled masks as binary images
3. display id of the objects that need to be labeled
4. move labeled data to 'labeled' directory


yuyu2172:
This PR no longer depends on `sh` module.

wkentaro:
nice

PR #1729, opened 2016-06-17
Stop doing self_filter while recognizing object in hand

wkentaro:
Because of not really calibrated kinect2 extrinsic parameters.


knorth55:
👍

PR #1733, opened 2016-06-18
add wait-opposite-arm when returning from bin

### LINKED ISSUE #1720: arms collide when picking from stow and return from bin
one arm: picking from tote
the other arm: return from bin

- [x] make wait_opposite_arm_return_from_bin

sometimes arms collide

knorth55:
Closes #1720


pazeshun:
+1

PR #1736, opened 2016-06-18
Install axis_camera from source because of its bug

wkentaro:
Reported but not yet released: https://github.com/ros-drivers/axis_camera/issues/46


wkentaro:
For https://github.com/start-jsk/jsk_apc/blob/master/jsk_2016_01_baxter_apc/launch/in_hand_data_collection.launch#L105

PR #1714, opened 2016-06-16
run sib with kinect

yuyu2172:
When running sib with kinect, run `sib_kinect.lauch` instead of `setup_softkinetic.launch`.

PR #1747, opened 2016-06-19
Enable to use kinect in picking-with-sib.l

pazeshun:
Works with #1714


pazeshun:
Tested in real robot

knorth55:
:+1:

PR #1758, opened 2016-06-20
set boundary to  avoid arm from moving tote 

### LINKED ISSUE #1719: arms move tote when it pick objects from tote
- [x] set boundary for picking region

~~- [ ] use smaller attention clipper~~

knorth55:
Closes #1719

### DIFF
--- a/jsk_2016_01_baxter_apc/euslisp/jsk_2016_01_baxter_apc/baxter-interface.l
+++ b/jsk_2016_01_baxter_apc/euslisp/jsk_2016_01_baxter_apc/baxter-interface.l
@@ -59,7 +59,7 @@
     (sethash :offset-object-h _hard-coded-variables -10)
     (sethash :offset-gripper-bin-top _hard-coded-variables 10)
     (sethash :offset-gripper-bin-side _hard-coded-variables 40)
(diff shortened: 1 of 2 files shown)


jsk_2016_01_baxter_apc/euslisp/jsk_2016_01_baxter_apc/baxter-interface.l
```diff
@@ -259,6 +261,22 @@
                   (send bbox :header :frame_id)
                   (send bbox :pose)))
       bx))
+  (:get-movable-region-for-tote
+    (arm &key (offset (list 0 0 0)))
+    (let (cube)
+      (setq cube (send self :bbox->cube (gethash arm _order-bin-box)))
```
pazeshun: This requires that `:recognize-order-bin-box` is called before it is called. You should avoid bugs like following:
```
(unless
   (setq order-bin-box (gethash arm _order-bin-box))
   (ros::ros-error "[:get-movable-region-for-tote] No data about order bin box. Call :recognize-order-bin-box first.")
   (return-from :get-movable-region-for-tote nil))
```

pazeshun:
+1

PR #1761, opened 2016-06-20
Changes to use new vacuum cleaners

ban-masa:
Firmware for new circuit to control 220V vacuum cleaners.
I have to check this firmware and circuit with the new vacuum cleaners.
@pazeshun please review this.


pazeshun:
Related to #1723

pazeshun:
I think it is better to update https://github.com/start-jsk/jsk_apc/blob/master/jsk_2016_01_baxter_apc/launch/include/vacuum_gripper.launch in this PR.

pazeshun:
Also, it is better to make the name of this PR more general.

pazeshun:
We should change the threshold of pressure sensor, but I think this should be done in another PR.

pazeshun:
+1

Issue #1423, opened 2016-05-10
Collect data for RBO segmentation with objects in APC2016


wkentaro:
## Summary

Currently we only have data for objects in APC2015, so we need to collect data for APC2016.

yuyu2172:
# Update

Currently, Baxter does not have light on its arms. This is problematic because objects near the back of bins get shaded, and color property of them and those near the entrances differ. 
@pazeshun and I discussed about this issue and we plan to update Baxter to have lights (hopefully!).

For the time being, I would like to collect data only for those with objects near the entrances. After lights are installed, I will collect data with objects near the back.
In order to execute this plan, I will limit the number of objects in bins to 3 as I found that objects get shaded when there are more than 3 objects in bins.

wkentaro:
That sounds reasonable.

yuyu2172:
First, I would collect 60 images (60 * 3 label images).
Then, I would collect 60 images more (total of 120 * 3 label images).

yuyu2172:
### June 14th memo

Collected data for a reduced object set of class size 20 with torso kinect.

+ Total number of data collected: 82 mask images
+ Elapsed time 90 minutes. (it could have been done in 1.2~1.5x faster)

I will create a dataset with data collected today and a part of RBO dataset for objects that exist both in 2015 and 2016 object lists.
The total number of mask images would be 100.

yuyu2172:
# June 20th memo
Collected 156 images with astra camera.
Each image contain one object.

yuyu2172:
# June 21st memo

Done labeling 156 masks.


Score measured from APC

all features acc  0.914290836581      
all features acc_cls  0.776140757289  
all features mean_iu  0.704529296258  
all features fwavcc  0.845199640749   


trained only by color features acc  0.90797209778        
trained only by color features acc_cls  0.756061724512   
trained only by color features mean_iu  0.682628242601   
trained only by color features fwavcc  0.834465511799

yuyu2172:
# June 24th memo

Collected 43 image sets with each containing 3 objects.
Took about 1.5h to collect and 1h to label.

PR #1763, opened 2016-06-20
train script for 2016 RBO dataset

yuyu2172:
#1423 

### usage
```
python train_segmenter_2016.py
```

You need to update rbo_segmentation submodule to use this.


knorth55:
you already have .gitignore in data dir
https://github.com/start-jsk/jsk_apc/blob/master/jsk_apc2016_common/data/.gitignore

yuyu2172:
fixed

PR #1765, opened 2016-06-21
astra cameras produce pointclouds of shelf edges that are parallel to the ground

yuyu2172:
## what I have done
I fixed 1st element of rotation parameters (R of RPY??) so that pointclouds of bin is parallel to the ground.

I adjusted kiva_pod_base so that a mask for SIB looks good.

## left hand

left_box1

left_box2

## video

https://drive.google.com/file/d/0BzBTxmVQJTrGZGdOSzBDWkVqbFk/view?usp=sharing


yuyu2172:
Rotation of right hand camera is adjusted as well so that view on shelf looks good.


#### before 
screenshot 2016-06-21 12 50 15





before








#### after
after

screenshot 2016-06-21 12 52 14

yuyu2172:
# right hand
right_box1
right_box2

PR #1766, opened 2016-06-21
modify view hand pose for stow-task

knorth55:
@pazeshun 
this PR prevents arms collision you saw yesterday
https://drive.google.com/file/d/0B5DV6gwLHtyJRzY4MzZVNEdmTFk/view?usp=sharing


pazeshun:
+1
BTW, can Kinect see objects in hand after `(send *baxter* arm :move-end-pos #f(-50 0 180) :world)`?

PR #1784, opened 2016-06-23
Abort picking objects when IK to it fails

pazeshun:


pazeshun:
Tested in real robot

pazeshun:
Sorry for big diff, but most of that is changing indentation.

pazeshun:
I think `:try-to-pick-object` is too big, but I'll make another PR to solve that problem.

pazeshun:
Fixed

ban-masa:
+1

PR #1790, opened 2016-06-24
Remove :recognize-objects-in-bin in picking-with-sib.l

pazeshun:
With this PR, picking-with-sib.l works without kinect head.


ban-masa:
+1

PR #1791, opened 2016-06-24
Fix position of wait-interpolation-smooth

pazeshun:
`:wait-interpolation-smooth` needs `send *ri* :angle-vector` before itself.


ban-masa:
+1

PR #1792, opened 2016-06-24
stop using self filters

yuyu2172:
Currently, self filter stopped working. Without this PR, segmentation_in_bin.launch does not start.


pazeshun:
Now self filter works.
I don't know why...

yuyu2172:
good

PR #1799, opened 2016-06-24
Data collection program for segmentation in bin

wkentaro:
Data is collected here: https://drive.google.com/open?id=0B9P1L--7Wd2vQ0NfVmNTS1pnemM


yuyu2172:
Subscribe to bininfo.

wkentaro:
> Also, is the data saving format consistent with the current convention?

NO, because this is only for FCN.

wkentaro:
> I think that it is not burdensome to set json when launching data collecting script.

Putting objects in bin according to json file takes time.

wkentaro:
It is annotated like below:

image
label
label_viz

yuyu2172:
Automatic collection is nice

PR #1800, opened 2016-06-25
[WIP] Split recognition part and deciding direction part from :try-to-pick-object

pazeshun:
To change the order of movement easily


pazeshun:
Close for now

PR #1809, opened 2016-06-25
add stow_work_order_server node

knorth55:
- consider total objects volume in bin
- consider bin points
- consider object in hand volume


pazeshun:
I couldn't understand workorder code fully, but +1

pazeshun:
@wkentaro Please check work order node. I'm not confident of that.

PR #1816, opened 2016-06-26
Enable to change offset and overlook-angle

pazeshun:


pazeshun:
Sorry, forget about this.

PR #1818, opened 2016-06-26
fcn_node: subtract mean-rgb from input data before doing segmentation

yuyu2172:


wkentaro:
Please paste the result here. (with more than 3 candidate objects)

PR #1821, opened 2016-06-26
wait longer before starting to subscribe to sib result

yuyu2172:
Wait time is too short especially when an arm has moved from a previous position.


pazeshun:
+1

PR #1850, opened 2016-07-15
Reflect new dataset on install_dataset script

yuyu2172:
I added 30 data and combined them with previous data.

https://drive.google.com/drive/u/0/folders/0BzBTxmVQJTrGOUFPbkJWUGV3a3c


wkentaro:
please use jsk_data.download_data.

yuyu2172:
fixed

PR #1855, opened 2016-07-21
depth register works with explicit arg

yuyu2172:
Without this PR, cameras did not publish depth registered points.

FYI, sheeta is equipped with titan X. It can run all launch files by itself.


yuyu2172:
rviz_screenshot_2016_07_21-15_56_15

PR #1868, opened 2016-08-26
Adjust right gripper firmware to gripper-v4

pazeshun:
#1856


pazeshun:
This firmware was already written to arduino

k-okada:
これは機構が変わったの？

k-okada:
did you changed the gripper mechanism?or this is just a bug

pazeshun:
The gripper mechanism was changed. The pose of servo motor was inverted.

pazeshun:
Now poses of right gripper servo and left gripper servo are symmetrical.

PR #1875, opened 2016-09-17
Publish bin bounding boxes in baxter.launch

wkentaro:
This is useful because we can use baxter-interface.l without main.launch or main_stow.lauch.


pazeshun:
I think so, too

PR #1909, opened 2016-10-21
Update fc.rosinstall for version 2.0.0

wkentaro:


fc.rosinstall
```diff
@@ -1,60 +1,32 @@
...
-    uri: https://github.com/code-iai/iai_kinect2.git
+    local-name: FOR_LATEST_BAXTEREUS/baxter_interface
+    uri: https://github.com/wkentaro/baxter_interface.git
+    version: baxtereus-devel
+- git:
+    local-name: FOR_LATEST_BAXTEREUS/jsk_robot
+    uri: https://github.com/jsk-ros-pkg/jsk_robot.git
     version: master
+- git:
+    local-name: FOR_PCL1.8/octomap_mapping
```
knorth55: `FOR_PCL1.8/OctoMap/octomapping` is better, isnt it?

fc.rosinstall
```diff
@@ -1,60 +1,32 @@
...
+    uri: https://github.com/OctoMap/octomap_mapping.git
+    version: indigo-devel
+- git:
+    local-name: FOR_PCL1.8/perception_pcl
+    uri: https://github.com/ros-perception/perception_pcl.git
```
knorth55: same as above

fc.rosinstall
wkentaro: I'm not sure which should I use if
- https://github.com/wkentaro/baxter_interface.git
- RethinkRobotics/baxter_interface or wkentaro/baxter_interface

That's why I chosen this, and too long local-name breaks readability of wstool info command output.
So I decided to use current local-name. Any idea?

knorth55: no idea.
it seems that your opinion is also correct.
anyway, i follow your idea.

PR #1910, opened 2016-10-21
Move srv to common package to fix dependency graph

wkentaro:
- dependency graph should be jsk_2016_01_baxter_apc -> jsk_apc2016_common

cc @knorth55


knorth55:
👍

knorth55:
thank you. i didnt think enough about it

PR #1917, opened 2016-10-22
compress debug log in rostest

knorth55:
- comment out irtviewer drawing function
- set debug-view nil


knorth55:
@pazeshun is it ok for this change? I thought you often used this test.

pazeshun:
+1
I sometimes used this, but I don't need viewer

Issue #1904, opened 2016-10-19
add two arm stow methods


knorth55:
for @knorth55's bthesis
- [x] add look opposite hand method
- [x] add hold object with two arm method
- [x] add putting object in bin with two arms method
- [x] detect object holding by two arms (`:wait-interpolation-until-grasp` method)

knorth55:
:innocent:

PR #1923, opened 2016-10-27
add :wait-interpolation-until-grasp method

knorth55:
related to #1904 

add methods to cancel angle-vector when arm detect grasping

### DIFF
--- a/jsk_2016_01_baxter_apc/euslisp/jsk_2016_01_baxter_apc/baxter-interface.l
+++ b/jsk_2016_01_baxter_apc/euslisp/jsk_2016_01_baxter_apc/baxter-interface.l
@@ -1392,6 +1392,13 @@
       (pushback (send *baxter* arm :inverse-kinematics coords-tmp :rotation-axis t) avs)
       (send self :angle-vector-sequence avs :fast nil 0 :scale 4.0)
       ))
+  (:wait-interpolation-until-grasp
+    (arm)
+    (while (send self :interpolatingp)
+           (if (send self :graspingp arm)
+             (progn
+               (ros::ros-info "[:wait-interpolation-until-grasp] Grasping detected. Cancel angle vector: ~a" arm)
+               (send self :cancel-angle-vector)))))
   (:place-object
     (arm)
     (send self :move-arm-body->order-bin arm)


jsk_2016_01_baxter_apc/euslisp/jsk_2016_01_baxter_apc/baxter-interface.l
```diff
@@ -1392,6 +1392,14 @@
...
       (send self :angle-vector-sequence avs :fast nil 0 :scale 4.0)
       ))
+  (:wait-interpolation-until-grasp
+    (arm)
+    (while (send self :interpolatingp)
+           (if (send self :graspingp arm)
+             (progn
+               (ros::ros-info "[:wait-interpolation-until-grasp] Grasping detected. Cancel angle vector: ~a" arm)
+               (send self :cancel-angle-vector))))
+    (send *baxter* :angle-vector (send self :state :potentio-vector)))
```
pazeshun: I think `(send *baxter* :angle-vector (send self :state :potentio-vector)))` is not needed because `:wait-interpolation-until-grasp` is the method just for `*ri*`

pazeshun: I think `(send *baxter* :angle-vector (send self :state :potentio-vector)))` is not needed because `:wait-interpolation-until-grasp` is the method just for `*ri*`

pazeshun: I think `(send *baxter* :angle-vector (send self :state :potentio-vector)))` is not needed because `:wait-interpolation-until-grasp` is the method just for `*ri*`

pazeshun:
Nice feature!
The reason why I use `:wait-interpolation-smooth` in pick task is to prevent crumpling of the vacuum pad.
We can start vacuuming before the arm finishes approaching to object, so the vacuum pad can grasp object without pushing it away when vacuuming force is weak enough.

This method can replace unreadable waiting for the end of approaching in https://github.com/start-jsk/jsk_apc/blob/master/jsk_2016_01_baxter_apc/euslisp/jsk_2016_01_baxter_apc/baxter-interface.l#L923.

pazeshun:
And with above thought, I think timeout is needed for grasp failure.

jsk_2016_01_baxter_apc/euslisp/jsk_2016_01_baxter_apc/baxter-interface.l
pazeshun: Your opinion is correct, but that behavior is totally different from `:wait-interpolation`.
I want `:wait-interpolation-until-grasp` to be as similar to `:wait-interpolation` as possible in order to avoid confusing.

knorth55: tell me more about the behavior of `smooth` one.
do they have gap between `*ri*` and `*baxter*`?

pazeshun: Yes.
```
12.irteusgl$ send *ri* :state :potentio-vector
#f(2.19727 45.1538 -4.74609 -55.2832 88.3081 76.2231 -32.5635 8.04199 0.0 -69.8291 -6.96533 82.9248 120.322 11.1401 6.17432 -108.149 0.0)
13.irteusgl$ send *baxter* :fold-pose-back
#f(0.0 97.4707 -2.39502 -94.5483 134.67 91.4062 8.70117 0.0 0.0 -97.4707 -2.39502 94.5483 134.67 -91.4062 8.70117 0.0 0.0)
14.irteusgl$ progn (send *ri* :angle-vector (send *baxter* :angle-vector) 10000) (send *ri* :wait-interpolation-smooth 5000) (send *ri* :state :potentio-vector)
#f(2.15332 70.8618 -4.15283 -72.9272 111.248 83.8916 -11.8433 4.06494 0.0 -83.606 -5.07568 87.0996 127.529 -40.144 8.50342 -54.0527 0.0)
15.irteusgl$ send *ri* :state :potentio-vector   ;; after the motion finished
#f(2.19727 97.4487 -2.37305 -94.4165 134.604 91.3843 9.20654 0.0 0.0 -97.4927 -2.43896 93.4937 134.67 -91.4502 9.14062 0.0 0.0)
```

knorth55: ok, i got it.
But do you know the reason why it is designed like that?

knorth55: What is  ptos and cons for it?
If synchronizing angle-vector is better, you can make PR for pr2eus.

pazeshun: I think real robot and `*baxter*` should be independent. Because we sometimes use `*baxter*` for some simulation without moving robot.
In pr2eus, real robot is equal to `*ri*`, so synchronization of angle vector is not good idea.

knorth55: hmm...
it is not what i want to know, but i got the reason.

critical example is as below.
we can keep angle-vector by `variable` and apply it while `*baxter*`'s angle-vecter is totally different.
then, when we use `wait-interpolation-until-grasp`, which synchronized `*baxter*` and `*ri*`, `*baxter*`'s posture will be overwritten.

```
$ (setq av (send *baxter* :angle-vector))
$ (send *baxter* :fold-pose-back)
$ (send *ri* :angle-vector av)
$ (send *ri* :wait-interpolation-until-grasp)
```

pazeshun: ごめんなさい。時間がないのでこれでいいです。
時間が経ってから見返すと、`:wait-interpolation` と違って`*baxter*`が勝手に変更されちゃうから、使用方法を間違えそうだなと思っただけです。

pazeshun:
+1

jsk_2016_01_baxter_apc/euslisp/jsk_2016_01_baxter_apc/baxter-interface.l
knorth55: いや、はせしゅんの指摘通りに変更する予定です。大丈夫です。

knorth55:
@pazeshun fixed, thx :innocent:

pazeshun:
Thanks! I'll apply this to pick task.

PR #1927, opened 2016-11-02
comment out vgg object verification node in main_stow.launch

knorth55:


knorth55:
@wkentaro ping

wkentaro:
Please add label.

PR #1928, opened 2016-11-04
[WIP] modify gripper arduino to pub FluidPressure msg

knorth55:
@pazeshun can you review this?
i tried to modify gripper arduino program and i got this error...
the error message says there is mismatch between ros version, but actually there is not.
the main problem is that the topic doesn't came back, and it goes to timeout.
(this error msg occurs timeout problem)

```
[ERROR] [WallTime: 1478271239.546674] Unable to sync with device; possible link problem or link software version mismatch such as hydro rosserial_python with groovy Arduino
```

### DIFF
--- a/jsk_2016_01_baxter_apc/firmware/gripper_front_manager_left/gripper_front_manager_left.ino
+++ b/jsk_2016_01_baxter_apc/firmware/gripper_front_manager_left/gripper_front_manager_left.ino
@@ -2,9 +2,9 @@
 #include <ros.h>
 #include <std_msgs/String.h>
 #include <std_msgs/Bool.h>
-#include <std_msgs/Float64.h>
 #include <std_msgs/Float32.h>
 #include <std_msgs/Int16.h>
+#include <sensor_msgs/FluidPressure.h>
 #include <Servo.h>
 
 const float pi = 3.14159;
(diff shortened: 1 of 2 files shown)


pazeshun:
I think this PR has no problem. And that error always occurs if connection to arduino has some problem (not only ROS version difference).
First of all, you should confirm whether writing firmware to arduino is successful. Did you write correct firmware to each arduino?
If there is no problem, I suspect memory shortage.

pazeshun:
I don't want arduino nano to handle heavy message in order to avoid buffer shortage problem.
For arduino nano, message which has header is very heavy (i.e. sensor_msgs).
Is this PR really needed? If you think so, I'll debug.

pazeshun:
Perhaps, you should `#include <ros/time.h>`

knorth55:
> First of all, you should confirm whether writing firmware to arduino is successful. Did you write correct firmware to each arduino?

i could compile and upload arduino program successfully.
the reason why i ask you is you are the guy who knows rosserial_arduino most in this team.

this PR add only header, which is not so big, and, in general, pressure msg should contain time information because it changes frequently.
however, there is no buffer and there is no solution, it doesn't bother me.

pazeshun:
Ok. I'll search what is the problem.
Maybe we can solve buffer shortage by changing default settings, but I couldn't find good settings in previous googling.

knorth55:
@pazeshun 
it's my problem, so do not spend your time much.
I just want to ask you if you already experienced this kind of problem.
now i'm thinking that rosserial does not work properly with sensor_msgs...

knorth55:
i tried with `#include <ros/time.h>` but it does not work

knorth55:
i googled a bit but it looks there is no enough buffer. 
because of buffer shortage of arduino nano, close this PR

knorth55:
memo:

i tried to enlarge buffer and i found that ATmega328P can be below
https://github.com/ros-drivers/rosserial/blob/jade-devel/rosserial_arduino/src/ros_lib/ros.h#L43-L51

```
ros::NodeHandle_<ArduinoHardware, 25, 25, 280, 280> nh;
```

Issue #1893, opened 2016-10-11
gripper_pad is not loaded to gazebo


knorth55:
Require https://github.com/start-jsk/jsk_apc/pull/1892 and https://github.com/RethinkRobotics/baxter_simulator/pull/88

Gazebo simulator cannot load gripper pad.

![gazebo_apc_tuck](https://cloud.githubusercontent.com/assets/9300063/19262672/49e12118-8fd3-11e6-8236-54f154066561.png)
![gazebo_gripper_pad](https://cloud.githubusercontent.com/assets/9300063/19262673/4a162804-8fd3-11e6-9398-0542e3c5db3c.png)

knorth55:
@pazeshun 
regenerate this condition
```
# baxter_simulator
roscd baxter_simulator
git remote add knorth55 https://github.com/knorth55/baxter_simulator.git
git fetch knorth55
git checkout knorth55/gripper-sim

# jsk_apc
roscd jsk_2016_01_baxter_apc
git remote add knorth55 https://github.com/knorth55/jsk_apc.git
git fetch knorth55
git checkout knorth55/fix-sim-apc2016

# roslaunch
roslaunch jsk_2016_01_baxter_apc baxter_pick_sim.launch
```

pazeshun:
@knorth55 Thank you.
I'll try to fix this.

knorth55:
#1940 i added gazebo tag, but it does not solve the problem.

knorth55:
#1975 solve to load robot model, but not colored

pazeshun:
I think http://gazebosim.org/tutorials/?tut=ros_urdf#NoteOnUnits will fix the problem

knorth55:
Totally Fixed via #1975 
![baxter_gazebo_load](https://cloud.githubusercontent.com/assets/9300063/20961437/563935f2-bca8-11e6-9620-5d86d4b9ebea.png)

PR #1937, opened 2016-11-10
add gripper_trajectory_server for simulator

knorth55:
without this node, Baxter in simulator does not move with roseus because of error below.

```
[ERROR] [1478767017.396381972, 20.989000000]: Could not start controller with name left_gripper_controller because no controller with this name exists
[ERROR] [1478767017.396454711, 20.989000000]: Could not start controller with name right_gripper_controller because no controller with this name exists
[ERROR] [1478767017.396497205, 20.989000000]: Failed to switch controllers
[ERROR] [1478767017.396525654, 20.989000000]: Failed to switch controllers
```

this gripper_trajectory_server is a dummy and do nothing for gazebo because gripper joint is not loaded in gazebo.

gazebo_eus

### DIFF
--- a/jsk_2016_01_baxter_apc/launch/include/baxter_sim.xml
+++ b/jsk_2016_01_baxter_apc/launch/include/baxter_sim.xml
@@ -109,4 +109,7 @@
         pkg="jsk_2015_05_baxter_apc" type="initialize_baxter.py"
         output="screen" />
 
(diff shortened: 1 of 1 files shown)


pazeshun:
+1
I need to add new `gripper_trajectory_server` after #1893 is fixed.

knorth55:
please merge this

PR #1939, opened 2016-11-11
add jskbaxter2 moveit config package

knorth55:
MERGE AFTER #1938 
related to #1936 

simulation: https://drive.google.com/open?id=0B5DV6gwLHtyJdXY1NDBfT0tMa3c
simulation with pillar: https://drive.google.com/open?id=0B5DV6gwLHtyJWGh1T0M1bkE0SEE
real robot: https://drive.google.com/open?id=0B5DV6gwLHtyJOGVEWExUVHM0bWc

PR #1941, opened 2016-11-14
add jskbaxter2 moveit_config

knorth55:
Not package version of #1939 

@k-okada how about this?


k-okada:
LGTM、let's go forward with this and see what happens.

PR #1942, opened 2016-11-14
add moveit-environment in baxter-interface

knorth55:
merge after #1941 or #1939 

default: moveit is disabled.
in case you want to use moveit, you need to set key `:moveit` `t`

```
(jsk_2016_01_baxter_apc::baxter-init :moveit t)
```

### DIFF
--- a/jsk_2016_01_baxter_apc/euslisp/jsk_2016_01_baxter_apc/baxter-interface.l
+++ b/jsk_2016_01_baxter_apc/euslisp/jsk_2016_01_baxter_apc/baxter-interface.l
@@ -31,7 +31,11 @@
 
 (defmethod jsk_2016_01_baxter_apc::baxter-interface
   (:init
-    (&rest args)
+    (&rest args &key
+           ((:moveit-environment mvit-env)
+            (instance jsk_2016_01_baxter_apc::baxter-moveit-environment))
+           ((:moveit-robot mvit-rb) (instance jsk_2016_01_baxter_apc::baxter-robot :init))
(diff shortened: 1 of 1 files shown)


pazeshun:
+1

jsk_2016_01_baxter_apc/euslisp/jsk_2016_01_baxter_apc/baxter-interface.l
```diff
@@ -90,6 +90,9 @@
       )
     (ros::advertise "/gripper_front/limb/left/servo/torque" std_msgs::Bool)
     (ros::advertise "/gripper_front/limb/right/servo/torque" std_msgs::Bool)
+    (setq moveit-robot (instance jsk_2016_01_baxter_apc::baxter-robot :init))
```
k-okada: global variable?

knorth55: it is in slots
https://github.com/jsk-ros-pkg/jsk_robot/blob/master/jsk_baxter_robot/baxtereus/baxter-interface.l#L14

knorth55:
updated
- set default `:moveit` `nil`

jsk_2016_01_baxter_apc/euslisp/jsk_2016_01_baxter_apc/baxter-interface.l
k-okada: I see

PR #1943, opened 2016-11-14
Change jsk_apc branch to gripper-v5

pazeshun:
This is for wstool.
I'll remove this commit when gripper-v5 is merged to master.

PR #1944, opened 2016-11-14
Add dynamixel-motor package to fc.rosinstall

pazeshun:
Depends on #1943 
`gripper-v5-devel` branch in pazeshun/dynamixel_motor includes arebgun/dynamixel_motor#67 and arebgun/dynamixel_motor#68. Without those PRs, we can't use Multi-turn mode of Dynamixel motor from dynamixel_motor package.


knorth55:
+1

PR #1947, opened 2016-11-16
Replace with refined KivaPod model

knorth55:
original repos: https://github.com/mDibyo/apc
original url: https://github.com/mDibyo/apc/blob/master/data/meshes/cubbyholes/pod_lowres.stl
original license: BSD

- this model, we can import kiva as scene in rviz

rviz_moveit_pod

@wkentaro how should i write license?


wkentaro:
My question is

- Why current kiva model is not usable?
- Why we need shelf model? Are box + shelf bin boxes not enough?

knorth55:
> - Why current kiva model is not usable?

current mesh has a lot of defects so that it cannot load to rviz.

> - Why we need shelf model? Are box + shelf bin boxes not enough?

i think, for moveit usage, it is easier to load this model than make scenes by yourself.

wkentaro:
> current mesh has a lot of defects so that it cannot load to rviz.

I see.

> this is for moveit scene.
i think, for moveit usage, it is easier to load this model than make scenes by yourself.

I mean we can convert the bin boxes published as `/publish_bin_boxes/output` to create the moveit scene,
and I thought this model is useless for moving arm to a specific bin.
If you're trying to replace the shelf box as manually defined below, for moving arm for `reset-pose` or `fold-pose-back`, it looks good for me.
https://github.com/wkentaro/jsk_apc/blob/fetch-moveit-apc/jsk_apc2016_common/config/transformable_shelf.yaml
https://github.com/wkentaro/jsk_apc/blob/fetch-moveit-apc/jsk_apc2016_common/moveit_interface.py#L27

knorth55:
> and I thought this model is useless for moving arm to a specific bin.

i don't get the reason why it is useless.
i think we can use it for moveit planning in stow task, like move arm from tote to bin-entrance.
Publish this model as scene, and euslisp solve ik and moveit do collision check and path planning.

wkentaro:
Is there the frame for each shelf bin?
For example bin_a_entrance_link?

2016年11月16日(水) 13:48 Shingo Kitagawa <notifications@github.com>:

knorth55:
no, but we can make as `publish_bin_boxes`
my main point is path planning and collision check from tote to bin-entrance.
in your code, you publish all bin boxes and check collision when robot approach to bin entrance?

BTW,
i thought BOX scene has no entrance arm cannot enter in because of collision, but is it correct?
https://github.com/wkentaro/jsk_apc/blob/fetch-moveit-apc/jsk_apc2016_common/moveit_interface.py#L39-L43

knorth55:
@wkentaro  ping

wkentaro:
I mean I can't understand its effectiveness without your showing some applications or experiments.

knorth55:
i thought to use this model to detect collision when robot arm move from tote to bin entrance.
however, it can be done by `moveit_commander.add_ box(~~~)` and mesh can be simplified in that case.
My current main purpose of this PR is just refine stl model.
Existing stl models is totally broken and malformed.

wkentaro:
+1

PR #1948, opened 2016-11-21
Add gripper-v5 model

pazeshun:
~~Note: I need to make Eus model like #1662 , or build may fail.~~ Done

PR #1949, opened 2016-11-21
Add ros_control nodes

pazeshun:
MERGE AFTER #1946


pazeshun:
Without #1948 , `joint_trajectory_controller_spawner` will fail due to the URDF, but it's ok

pazeshun:
This failure will be fixed in #1948

pazeshun:
Tested in real robot

PR #1954, opened 2016-11-24
Add eus methods to command hand and gripper

pazeshun:
* With this PR, we can command `*baxter*`'s hand like following:
```
5.irteusgl$ send *baxter* :hand :rarm :angle-vector #f(90 90)
(90.0 90.0)
```
screenshot from 2016-11-24 13 27 37
```
6.irteusgl$ send *baxter* :hand :rarm :angle-vector #f(0 0)
(0.0 0.0)
```
screenshot from 2016-11-24 13 28 19
* Also, we can get current angle
```
7.irteusgl$ send *baxter* :hand :rarm :angle-vector #f(10 20)
(10.0 20.0)
8.irteusgl$ send *baxter* :hand :rarm :angle-vector
(10.0 20.0)
```

* We can command real hand like following:
```
4.irteusgl$ send *ri* :move-hand :rarm (send *baxter* :hand :rarm :angle-vector) 1000 :wait t
#<control_msgs::followjointtrajectoryresult #X6f65410>
```

PR #1956, opened 2016-11-24
Add an example of picking with gripper-v5

pazeshun:
MERGE AFTER #1954


pazeshun:
TODO: Grasp paper towel

pazeshun:
Completed

PR #1966, opened 2016-11-29
Add an example of simple grasping with gripper-v5

pazeshun:


pazeshun:
I'll test this with #1973

PR #1969, opened 2016-11-29
Add load limit to gripper_ros_control

pazeshun:
This improves #1963 . #1963 is not enough.


pazeshun:
I'll add waning message

pazeshun:
Tested in real robot with #1968 and #1966

PR #1971, opened 2016-11-29
update baxter_common repo in fc.rosinstall

knorth55:
this repo is for jsk_apc.
Make collision head smaller than before.
It makes moveit to solve path successfully.

### DIFF
--- a/fc.rosinstall
+++ b/fc.rosinstall
@@ -37,4 +37,4 @@
 - git:
     local-name: FOR_SIMULATION/baxter_common
     uri: https://github.com/knorth55/baxter_common.git
-    version: if-gripper
+    version: jsk_apc


pazeshun:
+1
But I think we had better solve this problem by settings in move-it

knorth55:
collision head in `baxter_description` is too big and already malformed.
In moveit, it can ignore collision of `collision head`, but i thought it is better to make them smaller.
collision_head

pazeshun:
What kind of path can move-it solve thanks to this PR?
Could you  paste screenshot?

knorth55:
ik-bin-entrance :b

pazeshun:
I understood. Thanks

knorth55:
please merge this

PR #1972, opened 2016-12-02
fix trajectory_execution namespace

knorth55:
trajectory_execution namespace is wrong.
see https://github.com/ros-planning/moveit/issues/61


knorth55:
please merge this 😆

PR #1973, opened 2016-12-02
Realize stable grasping

pazeshun:
To grasp object with certain force


pazeshun:
Tested enough with real robot with #1956

PR #1974, opened 2016-12-05
[WIP] Add jsk_pr2eus package to fc.rosinstall

pazeshun:
To fix https://github.com/jsk-ros-pkg/jsk_robot/issues/738

### DIFF
--- a/fc.rosinstall
+++ b/fc.rosinstall
@@ -23,6 +23,10 @@
     uri: https://github.com/jsk-ros-pkg/jsk_robot.git
     version: master
 - git:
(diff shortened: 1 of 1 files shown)


pazeshun:
I need this regardless of how to fix because I have to change the behavior of `:angle-vector-sequence` anyway.

knorth55:
@pazeshun this branch is master, and you are waiting for release? I think we do not need to update this fc.rosinstall for waiting for release

pazeshun:
Sorry, perhaps this is not needed.

pazeshun:
We don't need this because the problem will be fixed in https://github.com/wkentaro/baxter_interface/pull/1

PR #1984, opened 2016-12-16
Adjust picking-with-sib to current object segmentation

pazeshun:
To follow 2c768428141ee7c9e4ad16f732688f91dc7c145a and 178240324e56606eb98b1dd3e8196d77c3d7abf4


pazeshun:
Tested with both gripper-v4 and gripper-v5

PR #1985, opened 2016-12-16
Don't move gripper while carrying object in picking-with-sib

pazeshun:


pazeshun:
Tested with both gripper-v4 and gripper-v5

pazeshun:
Ping @wkentaro

wkentaro:
Please assign me to request the review, and add description of what is this and why this is needed.

pazeshun:
@wkentaro 
The base of this PR is master, so I want you to merge this.
But the description about this PR is too few, so I'll explain.

What I want to do is not to move grippers while carrying objects in picking-with-sib.l
So I added `no-gripper-controller`. By using this controller like `(send *ri* :angle-vector (send *baxter* :angle-vector) 3000 :no-gripper-controller)`, you can move arms and head without moving gripper joints.
I also changed `:send-av` and `:move-arm-body->order-bin`. You can set controller to these methods to select which part of robot moves.
Finally, I set `no-gripper-controller` to `:send-av` and `:move-arm-body->order-bin` in picking-with-sib.l

I checked these changes have backward compatibility. And I think these changes are useful to both gripper-v4 and gripper-v5.
So I send this PR to master

PR #1993, opened 2016-12-31
Fix :ik->nearest-pose not to move arms unintendedly

pazeshun:
### What is problem
* When all IK fails, `:ik->nearest-pose` returns nil, but forgets to revert `*baxter*`'s angle-vector to the one before `(send *baxter* :angle-vector pose)`
* The opposite arm moves after `:ik->nearest-pose`


pazeshun:
@knorth55 Too long, isn't it?

pazeshun:
I'll fix current-angle-vector to current-av and opposite-av to current-opposite-av

pazeshun:
Fixed

knorth55:
Thats much better

wkentaro:
Please rebase.

pazeshun:
rebased

pazeshun:
@wkentaro Please merge this

PR #1994, opened 2016-12-31
Add :rotate-wrist-ik

pazeshun:
## What is problem
I feel IK failures in picking task are sometimes caused by wrist angle limit.

I'll show you one example:
Baxter's initial pose
screenshot from 2017-01-01 00 51 45
got by
```
36.irteusgl$ send *ri* :ik->bin-entrance :larm :b :offset #f(-30 0 0) :rotation-axis t
```
I want to solve IK to the coordinate in the picture above:
```
37.irteusgl$ ec
#<coordinates #Xa1ab078  965.957 13.486 844.095 / 0.034 -0.056 1.566>
```
Normal IK fails:
```
38.irteusgl$ send *baxter* :larm :inverse-kinematics ec
; failed for normal ik, starting from relaxed position
; failed for normal ik, try to move arms very slowly
; failed for slow ik, try to start from prepared poses
; starting from prepared pose ':untuck-pose'
nil
```
This is caused because wrist angle is too near joint limit:
```
39.irteusgl$ send *baxter* :larm :wrist-r :joint-angle
115.831
40.irteusgl$ send *baxter* :larm :wrist-r :max-angle
175.268
```
But wrist has wide range to opposite direction, too.
```
42.irteusgl$ send *baxter* :larm :wrist-r :min-angle
-175.268
```
So rotating wrist to opposite direction is the solution.
```
45.irteusgl$ send *baxter* :larm :wrist-r :joint-angle -180 :relative t
-64.1695
46.irteusgl$ send *baxter* :larm :inverse-kinematics ec
#f(0.0 -34.2761 -66.8794 -76.907 69.884 -70.546 -61.3844 -166.702 90.0 -20.0 -25.0 40.0 60.0 20.0 80.0 0.0 0.0)
```
screenshot from 2017-01-01 00 52 03
`(send *ri* :send-av 1000)` will move robot like below:
wrist_rotating

## What I did
In `:rotate-wrist-ik` method, I set 6 patterns of angle vector whose wrist angle is different each other to `pose-candidates` of `:ik->nearest-pose`.
With this, you can ignore wrist angle limits:
```
48.irteusgl$ send *ri* :ik->bin-entrance :larm :b :offset #f(-30 0 0) :rotation-axis t
; failed for normal ik, starting from relaxed position
; failed for normal ik, starting from relaxed position
; failed for normal ik, try to move arms very slowly
; failed for normal ik, starting from relaxed position
; failed for normal ik, starting from relaxed position
[...]

PR #1995, opened 2017-01-01
Add script to check rotate-wrist-ik

pazeshun:
Depends on #1994 
In fact, #1994 doesn't have much effect on gripper-v4. On the other hand, it has good effect on gripper-v5, because I have to rotate gripper-v5 around world x axis very frequently.

This script imitates the possible motions of arm in `:pick-object` method with gripper-v5.
We can check whether there is strange behaviour by looking at the kinematic simulator.


pazeshun:
Moved to #1996

PR #1996, opened 2017-01-01
Add script to check motion of pick-object with rotate-wrist-ik

pazeshun:
Depends on #1994
In fact, #1994 doesn't have much effect on gripper-v4. On the other hand, it has good effect on gripper-v5 because I have to rotate gripper-v5 around world x axis very frequently in pick task.

This script imitates the possible motions of arm in :pick-object method with gripper-v5 using :rotate-wrist-ik.
We can check whether there is strange behaviour by looking at the kinematic simulator.


knorth55:
no idea for gripper-v5, so I remove myself from reviewer

pazeshun:
Oh, sorry...

pazeshun:
memo: I'll rebase gripper-v5 onto master after #1994 is merged. After that, I'll rebase this branch onto gripper-v5.

Issue #1997, opened 2017-01-06
Segmentation Fault on test code


pazeshun:
In travis, I got following errors:
```
[jsk_2015_05_baxter_apc:make] wrote test file to [/home/travis/ros/ws_jsk_apc/build/jsk_2015_05_baxter_apc/test_results/jsk_2015_05_baxter_apc/shell-test_roseus__home_travis_ros_ws_jsk_apc_src_jsk_apc_jsk_2015_05_baxter_apc_euslisp_test_test-move-arm-to-bin.l.xml]



                                                                                
[jsk_2015_05_baxter_apc:make] FAILURE:


                                                                                
[jsk_2015_05_baxter_apc:make] [33m;; loading roseus("") on euslisp((9.22 ip-172-31-3-45 Thu Dec 29 20:56:31 PST 2016  1.0.14))


                                                                                
[jsk_2015_05_baxter_apc:make] [0mTEST-NAME: test-move-arm-to-bin


                                                                                
[jsk_2015_05_baxter_apc:make]   now testing...


                                                                                
[jsk_2015_05_baxter_apc:make] [32mstart testing [test-move-arm-to-bin]


                                                                                
[jsk_2015_05_baxter_apc:make] [0mCall Stack (max depth: 20):


                                                                                
[jsk_2015_05_baxter_apc:make]   0: at (gethash bin *bin-cubes*)


                                                                                
[jsk_2015_05_baxter_apc:make]   1: at (gethash bin *bin-cubes*)


                                                                                
[jsk_2015_05_baxter_apc:make]   2: at ;; Segmentation Fault.
```
```
[jsk_2016_01_baxter_apc:make] [0m[33moutput to /home/travis/ros/ws_jsk_apc/build/jsk_2016_01_baxter_apc/test_results/jsk_2016_01_baxter_apc/rosunit-test_ik_in_bin.xml



                                                                                
[jsk_2016_01_baxter_apc:make] [0mCall Stack (max depth: 20):


                                                                                
[jsk_2016_01_baxter_apc:make]   0: at (gethash key *bin-cubes*)


                                                                                
[jsk_2016_01_baxter_apc:make]   1: at (gethash key *bin-cubes*)


                                                                                
[jsk_2016_01_baxter_apc:make]   2: at (setq cube (gethash key *bin-cubes*))



                                                                                
[...]

PR #1998, opened 2017-01-06
use alist to avoid segmentation fault

knorth55:
temporary solution for #1997
`*bin-cubes*` hashtable can not use `gethash` but `:list` so that use plist instead of hashtable


pazeshun:
@wkentaro Please merge this

PR #2000, opened 2017-01-18
Fix model of gripper-v5

pazeshun:
To put colors on pads and tube for easy viewing.
Eus model after this PR:
screenshot from 2017-01-18 23 14 46
screenshot from 2017-01-18 23 16 06
screenshot from 2017-01-18 23 16 26


pazeshun:
But unable to view colors on Rviz:
screenshot from 2017-01-18 23 23 16
I assume this is the bug of Rviz, but I have no time to debug.
For now, I only need refined eus model.

PR #2006, opened 2017-02-04
Fix 404 of trained data vgg16_rotation_translation_brightness_372000...

wkentaro:
image


wkentaro:
I'm not sure why, but maybe I unexpectedly changed the id of file on gdrive.

PR #2011, opened 2017-02-09
FCN32s-V2: Update fcn32s trained model

wkentaro:
- Trained with dataset v2 (JSK + MIT) > 5k
- 148000 iterations
- Dataset in shelf + tote.


knorth55:
:+1:

PR #2021, opened 2017-03-05
Fix rarm endcoords in baxter.yaml

wkentaro:
I'm wondering why translate values in baxter.yaml are different
among rarm and larm


pazeshun:
Is there any problem?

wkentaro:
I think there is difference between model endcoords and real one.
The robot fails to pick object because of the **y** offset difference.

k-okada:
BTW, convention on euslisp coordinates system is ;
- when you grab a cup
- z is up
- x is front

so that `(send *robto* :larm :inverse-kinematics (send *cup* :worldcoords)` works.

screenshot from 2017-03-06 09 56 52

PR #2044, opened 2017-03-30
[WIP] updating samples for ARC picking task using fetch

YutoUchimi:
We took a sample video of fetch picking an item from table.
Here is the link to video.

https://drive.google.com/open?id=0BxzzMEwOXXe9SXVZT1pHc1lCMDg


YutoUchimi:
moved to #2045

PR #2048, opened 2017-04-05
Use rqt_yn_btn instead of rviz plugin

wkentaro:
In order to avoid the blocking of topic update on rviz.


wkentaro:
@pazeshun sorry, for the lack of sharing information.

- The rviz plugin currently blocks rviz thread to get user input
- You need to push button **twice** because we have two arms, left and right ones.

pazeshun:
Oh, I see.

wkentaro:
?

pazeshun:
If you deleted button on Rviz, why jsk_rviz_plugin/YesNoButton is loaded (https://github.com/start-jsk/jsk_apc/blob/708b233579469ab9c4d6e630ad0a424b60eda9af/jsk_2016_01_baxter_apc/rvizconfig/pick.rviz#L29)?
And why YesNoButton isn't deleted on Window Geometry (https://github.com/start-jsk/jsk_apc/blob/708b233579469ab9c4d6e630ad0a424b60eda9af/jsk_2016_01_baxter_apc/rvizconfig/pick.rviz#L659)?

PR #2084, opened 2017-05-12
Add usage documentation of baxter for ARC2017

708yamaguchi:
add usage of arc-interface.l


knorth55:
can you make this doc in rst and add it in `jsk_apc/doc`?
we have doc site like this http://jsk-apc.readthedocs.io/en/latest/.
You can try doc in localhost as below.

```
roscd jsk_apc 
cd ../doc
sudo pip install -r requirement
make html && make auto
# access to http://127.0.0.1:8888/ in your web browser
```
Then, your change in local will be automatically reflected to http://127.0.0.1:8888/.
How to write rst doc is here http://docutils.sourceforge.net/rst.html.

knorth55:
Can you remove `readme.md`? Or do we need it?

PR #2086, opened 2017-05-14
Depends at test time on jsk_2016_01_baxter_apc

wkentaro:
Without this change, we have error of failed finding of jsk_2016_01_baxter_apc at `find_package` in test block.


knorth55:
It is my mistake.
Can you modify to find_package `jsk_apc2016_common` because `euslint` moves to the package.

PR #2091, opened 2017-05-16
Move astra_hand.launch from setup_for_pick.launch to baxter.launch

YutoUchimi:


knorth55:
Is this a bug?

wkentaro:
Yeah.

wkentaro:
Missing change in https://github.com/start-jsk/jsk_apc/pull/2085

knorth55:
Why do you move astra launch to baxter launch?

wkentaro:
Because I think it is useful for people who do not use `setup_for_pick.launch`.

knorth55:
camera launch should be included in robot launch, thats true.
jsk_2016 should be, too.

PR #2093, opened 2017-05-17
update midpose to go back fold-pose-back

knorth55:
fix https://github.com/start-jsk/jsk_apc/issues/2074#issuecomment-300992304


knorth55:
updated

PR #2097, opened 2017-05-17
fix default-controller not to set head-controller twice

knorth55:
This causes `(send *ri* :wait-interpolation)` does not finish when you use `:default-controller`.


jsk_2016_01_baxter_apc/euslisp/jsk_2016_01_baxter_apc/baxter-interface.l
```diff
@@ -110,11 +110,12 @@
   ;; Overwrite super class's :default-controller
   (:default-controller
     () ;; larm-controller include head-controller
```
pazeshun: Could you fix this comment?

knorth55: updated

knorth55:
updated

Issue #2111, opened 2017-05-22
Introduce gripper-v6 with pinching motion


pazeshun:
gripper-v5 + [flex sensor](https://www.switch-science.com/catalog/508/) + [proximity sensor](https://roboticmaterials.com/rm/product/single-sensor-evaluation-kit/) -> gripper-v6

pazeshun:
@pazeshun
* ハードウェア全体を作る
  * [ ] 右手組み立て
  * [ ] 左手部品購入
  * [ ] 左手組み立て
* 曲げセンサを用いたピッキング動作を作る

pazeshun:
@708yamaguchi 
* 近接センサを扱う回路とファームウェアを作る
* Fetchで近接センサを使ってみて、gripper-v6でもできそうな手探り動作を調べる

pazeshun:
今週中に、右腕にgripper-v6を着けた状態でpick_main.launchを動かせるように頑張ります

knorth55:
'This week' you mean until this Saturday?
I think it is too hard.
You should not be too rushed. Feel free to make an achievable goal for you.

wkentaro:
Right hand without pinching motion is done by 3.1.0

wkentaro:
Closed via https://github.com/start-jsk/jsk_apc/pull/2203

PR #2115, opened 2017-05-25
Add Arduino firmware for right gripper-v6

pazeshun:
For #2111


ban-masa:
http://www.musashinodenpa.com/arduino/ref/index.php?f=0&pos=2113
According to the above page, analogRead returns int value.
I think finger_flex_msg should be Int type.

pazeshun:
Fixed

ban-masa:
:+1:

PR #2118, opened 2017-05-27
set *.launch and *.test as XML files

knorth55:


wkentaro:
why you need this?

knorth55:
these files are written in XML, but github analyze them as HTML.

wkentaro:
Is this not enough for you? https://github.com/start-jsk/jsk_apc/search?l=xml

wkentaro:
It is now fixed by Github.

PR #2148, opened 2017-06-14
[WIP] Allow moveit is not set

pazeshun:


knorth55:
can you add it in `lib/pick-interface.l`?
like 
```
(when (boundp `*co*) (send *ti* :add-shelf-scene))
```

knorth55:
Or I can do it by myself

pazeshun:
> can you add it in lib/pick-interface.l?
> like
> 
> (when (boundp `*co*) (send *ti* :add-shelf-scene))

Ok, I'll do this.

knorth55:
move to #2149

PR #2138, opened 2017-06-10
Add stow task and update pick task

knorth55:
Related to #2110 and #2053 

# How to use

```bash
roslaunch jsk_arc2017_baxter baxter.launch pick:=false
roslaunch jsk_arc2017_baxter setup_for_stow.launch
roslaunch jsk_arc2017_baxter stow.launch
```

# Main Change
- add stow_task files in `jsk_arc2017_baxter`
  - `euslisp/stow-main.l`
  - `euslisp/lib/stow-interface.l`
  - `launch/setup/setup_for_stow.launch`
  - `launch/main/stow.launch`
  - `rvizconfig/stow.rviz`
- add stow_task files in `jsk_arc2017_common`
  - `scripts/stow_json_generator.py`
  - `config/stow_task/shelf_marker.yaml`
  - `data/json/sample_stow_task`
- `node_scripts/state_server.py` support stow_task
- remove slot `bin-boxes-` and replace by `bin-cubes-` in `euslisp/lib/arc-interface.l`
- generalize picking methods for pick task and stow task in `euslisp/lib/arc-interface.l`
- add stow task methods and slots in `euslisp/lib/arc-interface.l`

# Minor Change
- remove method `:ik->bin-entrance` and replace by `ik->bin-center`
- rename method `:recognize-objects-in-bin` -> `:recognize-target-object` in `arc-interface.l`
- rename method `:ik->cardboard-entrance` -> `:ik->cardboard-center`
- reset slot `order` in `pick-interface.l`
- fix typo in `pick_json_generator.py`
- update pick state machine in `pick-main.l`

# State Machine 
## Pick Task
Updated!
arc_pick_state_machine

## Stow Task
arc_stow_state_machine

# Video
https://drive.google.com/open?id=0B5DV6gwLHtyJbkRmbkZxZEhXVjg

PR #2152, opened 2017-06-14
Fix small bugs added when adding stow

pazeshun:
Fix for #2138 
I also revert offsets for bin overlook pose

### DIFF
--- a/jsk_arc2017_baxter/euslisp/lib/arc-interface.l
+++ b/jsk_arc2017_baxter/euslisp/lib/arc-interface.l
@@ -361,9 +361,9 @@
       (send *baxter* :reset-pose arm)
       (send *baxter* :rotate-gripper arm gripper-angle :relative nil)
       (setq offset
-            (cond ((eq bin :a) (float-vector 0 offset-y 250))
-                  ((eq bin :b) (float-vector 0 offset-y 200))
-                  ((eq bin :c) (float-vector 0 offset-y 200))))
+            (cond ((eq bin :a) (float-vector -100 offset-y 250))
+                  ((eq bin :b) (float-vector -200 offset-y 200))
+                  ((eq bin :c) (float-vector -150 offset-y 200))))
       (setq rpy
             (float-vector 0 pi/2 (if (eq arm :larm) pi/2 -pi/2)))
       (pushback
--- a/jsk_arc2017_baxter/euslisp/lib/arc-interface.l
+++ b/jsk_arc2017_baxter/euslisp/lib/arc-interface.l
@@ -196,7 +196,7 @@
          (setq is-recognized nil)))
       is-recognized))
   (:pick-object-in-bin
-    (arm &rest args)
+    (arm bin &rest args)
     (let (graspingp movable-region)
       (setq movable-region (gethash bin bin-movable-regions-))
       (unless movable-region
--- a/jsk_arc2017_baxter/euslisp/lib/pick-interface.l
+++ b/jsk_arc2017_baxter/euslisp/lib/pick-interface.l
@@ -130,8 +130,7 @@
     ;; TODO: in-hand verification
     ;; TODO:   target-obj = in-hand object  -> :set_target_cardboard
     ;; TODO:   target-obj != in-hand object -> :return_object
-    (setq target-obj t)
-    target-obj)
+    t)
   (:check-picking-fail-count (arm)
     (setq picking-fail-count (incf picking-fail-count))
     (ros::ros-info "[main] arm: ~a, picking fail count: ~a" arm picking-fail-count)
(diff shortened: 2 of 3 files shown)


jsk_arc2017_baxter/euslisp/lib/arc-interface.l
```diff
@@ -361,9 +361,9 @@
       (send *baxter* :reset-pose arm)
       (send *baxter* :rotate-gripper arm gripper-angle :relative nil)
       (setq offset
-            (cond ((eq bin :a) (float-vector 0 offset-y 250))
-                  ((eq bin :b) (float-vector 0 offset-y 200))
-                  ((eq bin :c) (float-vector 0 offset-y 200))))
+            (cond ((eq bin :a) (float-vector -100 offset-y 250))
+                  ((eq bin :b) (float-vector -200 offset-y 200))
+                  ((eq bin :c) (float-vector -150 offset-y 200))))
```
knorth55: this value is modified because i change to use `:ik->bin-center`.
It doesn't work?

jsk_arc2017_baxter/euslisp/lib/pick-interface.l
```diff
@@ -130,6 +130,7 @@
     ;; TODO: in-hand verification
     ;; TODO:   target-obj = in-hand object  -> :set_target_cardboard
     ;; TODO:   target-obj != in-hand object -> :return_object
+    (setq target-obj t)
```
knorth55: `target-obj` is already set in line 59.
you don't need to set.

jsk_arc2017_baxter/euslisp/lib/arc-interface.l
```diff
@@ -196,7 +196,7 @@
          (setq is-recognized nil)))
       is-recognized))
   (:pick-object-in-bin
-    (arm &rest args)
+    (arm bin &rest args)
```
knorth55: thx

knorth55:
thx!

jsk_arc2017_baxter/euslisp/lib/pick-interface.l
pazeshun: The return value of `:verify-object` is used for state machine.
If we set `target-obj` as object name, state machine dies.
In stow-interface, you set `t`:
https://github.com/start-jsk/jsk_apc/blob/master/jsk_arc2017_baxter/euslisp/lib/stow-interface.l#L104

knorth55: I think `smach` doesn't die because of it.
what you want is return `t`, so return `t`. 
Do not set `t` in `target-obj`.

jsk_arc2017_baxter/euslisp/lib/arc-interface.l
pazeshun: Yes, IK fails in bin b. 

I decided those values considering current `:ik->bin-center` position, old `:ik->bin-entrance` position and old offset.
Those values moves arm to the same position as old position.

knorth55: ok thx.

jsk_arc2017_baxter/euslisp/lib/pick-interface.l
pazeshun: > I think smach doesn't die because of it.

At least, 
```
$ roslaunch jsk_arc2017_baxter baxter.launch moveit:=true pick:=true
$ roslaunch jsk_arc2017_baxter setup_for_pick.launch
$ roslaunch jsk_arc2017_baxter pick.launch json_dir:=$(rospack find jsk_arc2017_common)/data/json/easy_pick_task moveit:=true
```
dies due to smach error without this change.
Currently, I can't write simple smach test due to lack of knowledge, so can you check?

knorth55: please put an error output.

knorth55: anyway, please do not set `t` in `target-obj`.
just return `t` in the method.

pazeshun: Ok, I'll fix

pazeshun: fixed

knorth55:
thx so much!

jsk_arc2017_baxter/euslisp/lib/pick-interface.l
knorth55: Now I got why I made this mistake.
I update the PR #2172 with this review and I forgot to test it.
Maybe you are right, and we need to return `t` or `nil` if we make a state-machine with `make-state-machine` function.
I'm so sorry about it.
https://github.com/start-jsk/jsk_apc/pull/2127#discussion_r119985031

pazeshun:
Please merge this

PR #2153, opened 2017-06-14
Use roseus_smach from source

pazeshun:
Temporary solution
https://github.com/jsk-ros-pkg/jsk_roseus/issues/524#issuecomment-308374783


knorth55:
We can wait its release.

wkentaro:
Have you requested?

pazeshun:
Requested in https://github.com/jsk-ros-pkg/jsk_roseus/issues/524

PR #2160, opened 2017-06-19
[WIP] Add pinching with gripper-v6

pazeshun:
Depends on #2155 
Use pinching of gripper-v6 in pick and stow task.


pazeshun:
## Current Movies

## Current Problems
- https://github.com/start-jsk/jsk_apc/pull/2155#issuecomment-308910205
  - [ ] Use prismatic joint load to detect pushing
- Pinching `:graspingp` is almost `t`
  - [ ] Fix right finger flex sensor
  - [ ] Adjust thresholds
- Pinching often fails because object moves from recognized position by gripper pushing in suction grasp
  - [ ] Split `pick-object` state into `suction-object` and `pinch-object`

pazeshun:
After I fix fingers, I want to:
- [ ] change pinching direction depending on side length of object bbox

pazeshun:
Fixes #2161

pazeshun:
I have to fix fingers again...
Please wait until this weekend...

pazeshun:
Move to #2203

PR #2166, opened 2017-06-22
Update pick_task methods for new task environment

knorth55:
## Main Changes
- update cardboard pos for new env
- update cardboard moveit methods
- update pick motion parameters for new env

### jsk_arc2017_baxter
- `config/pick_task/shelf_bin.yaml`
- `config/pick_task/shelf_marker.yaml`
### jsk_arc2017_common
- `scripts/pick_json_generator.py`
- `json/sample_pick_task

PR #2167, opened 2017-06-23
Add moveit path constraints for place object

knorth55:
Merge after #2166 

I haven't tested on real robot yet.


knorth55:
I tested on real robot.
Video: 
https://drive.google.com/open?id=0B5DV6gwLHtyJWDlDcld3MnNhcFU

PR #2168, opened 2017-06-23
add optional bilateral filter node

YutoUchimi:
- Added bilateral filter for /right_hand_camera/stereo/depth_registered/points.
It is optional (i.e. set in argument), because the result of
`rostopic hz /right_hand_camera/stereo/depth_registered/filtered_points`
is 10 ~ 15 hz, while
`rostopic hz /right_hand_camera/stereo/depth_registered/points`
is 28 ~ 30 hz.

- Calibrated extrinsic parameter of right hand camera.

- Tuned stereo_image_proc parameters with dynamic reconfigure.


wkentaro:
Please add some images.

wkentaro:
Please reopen when you need this again.

PR #2174, opened 2017-06-24
Calibrate extrinsic parameters of hand camera and improve stereo options

wkentaro:
Simplified version of https://github.com/start-jsk/jsk_apc/pull/2168


wkentaro:
@YutoUchimi Could you please add why you need the parameters of stereo?

PR #2183, opened 2017-06-29
Calibrate intrinsic parameter of left hand camera

YutoUchimi:


YutoUchimi:
before
screenshot from 2017-06-29 22 14 26

after
screenshot from 2017-06-29 23 02 31

Issue #2185, opened 2017-06-29
Visualize JSON file of item location


wkentaro:
- [x] Write object location to item_location_file.json @knorth55 
- [x] Implement visualization script for item_location_file.json @wkentaro

knorth55:
Write object location to item_location_file.json  -> #2188

PR #2188, opened 2017-06-29
save json in pick task

knorth55:
Related to #2185 

- `json_saver.py` supports pick task
- change `:update-json` API
- modify json save dir

## Usage

```lisp
;; tote -> shelf
(send self :update-json target-obj :src :tote :dst (cons :bin target-bin))
;; shelf -> cardboard
(send self :update-json target-obj :src (cons :bin target-bin) :dst (cons :cardboard target-cardboard))
;; shelf -> shelf
(send self :update-json target-obj :src (cons :bin target-bin) :dst (cons :bin target-bin))
```


knorth55:
update to save `order_file.json` in output dir.

PR #2199, opened 2017-06-30
introduce left stereo astra camera

knorth55:
thanks to @YutoUchimi and @pazeshun

@YutoUchimi can you review this?


wkentaro:
We also tested with real robot in stow task.

PR #2201, opened 2017-07-01
Calibrated extrinsic parameters of right_hand_stereo by @YutoUchimi

wkentaro:
fork of https://github.com/start-jsk/jsk_apc/pull/2168


YutoUchimi:
Actually the result of this calibration is not good enough to recognize 3-dimentional position of objects placed near the edge of the camera frame.
I will probably make another PR which improves the accuracy than this PR.

PR #2208, opened 2017-07-03
speed up finding largest object index

knorth55:


knorth55:
this PR does not affect well.
Input topics are 2.2hz and output is 0.3hz.
Connection doesn't seem to be a problem, and the process inside seems crucial.

wkentaro:
Have you open an issue for this?

PR #2215, opened 2017-07-04
Semantic segmentation of unknown objects

wkentaro:
Close https://github.com/start-jsk/jsk_apc/issues/2173

I trained using apc2016 objects as unknown object, and tested with apc2015 objects.

**before**

**after**

**visualization (validation)**

iter00072000


pazeshun:
This is just a question, but https://github.com/start-jsk/jsk_apc/issues/2173#issuecomment-311729614 is fixed?

wkentaro:
@pazeshun Yes, and I added comment to the issue: https://github.com/start-jsk/jsk_apc/issues/2173#issuecomment-312909489

PR #2233, opened 2017-07-07
Use transformable_markers_client to adjust scene

wkentaro:
image


wkentaro:
jsk_recognition and jsk_visualization are released.

PR #2236, opened 2017-07-07
Add sleep in :wait-interpolation-until loop

pazeshun:
https://github.com/start-jsk/jsk_apc/pull/2235#discussion_r126157616

### DIFF
--- a/jsk_arc2017_baxter/euslisp/lib/baxter-interface.l
+++ b/jsk_arc2017_baxter/euslisp/lib/baxter-interface.l
@@ -323,7 +323,8 @@
       (if (send self :graspingp arm)
         (progn
           (ros::ros-info "[:wait-interpolation-until-grasp] Grasping detected. Cancel angle vector: ~a" arm)
-          (send self :cancel-angle-vector)))))
+          (send self :cancel-angle-vector)))
+      (unix::usleep 1000)))
   (:wait-interpolation-until
     (arm &rest args)
     (when (send self :simulation-modep)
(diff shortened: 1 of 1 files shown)


jsk_arc2017_baxter/euslisp/lib/baxter-interface.l
```diff
@@ -375,7 +376,8 @@
       (while (send self :interpolatingp)
         (when (reduce #'(lambda (x y) (or x y)) (mapcar #'funcall conds))
           (ros::ros-info "[:wait-interpolation-until] Cancel angle vector: ~a" arm)
-          (send self :cancel-angle-vector)))))
+          (send self :cancel-angle-vector))
+        (unix::usleep 1000))))
```
knorth55: BTW, Do we really need these lines? lime373-375
```
;; wait for :interpolatingp
    (dotimes (x 100)
      (if (send self :interpolatingp) (return))
      (unix::usleep 1000))
```

knorth55:
I have no idea why you need to `unix::usleep`.
Write more detailed information for review.

knorth55:
Is this for waiting `cancel-angle-vector` execution?

jsk_arc2017_baxter/euslisp/lib/baxter-interface.l
pazeshun: Just after sending angle-vector, interpolatingp sometimes return nil even if robot moves correctly

knorth55: ok

pazeshun:
It's for reducing CPU load.
Without sleep, one CPU load becomes nearly 100%, even if we don't need such power to detect signal of 100 Hz

knorth55:
Ok now i understand.
Please write more detailed info if you request a review unless it is quite troublesome to find out the context.

pazeshun:
Sorry

PR #2252, opened 2017-07-11
add jsk_pr2eus in fc.rosintall

knorth55:
we need master branch of jsk_pr2eus


wkentaro:
Please write why you need this to fc.rosinstall

k-okada:
fyi: https://github.com/ros/rosdistro/pull/15477

PR #2261, opened 2017-07-12
exclude head-controller from [l/r]arm-controller and add [l/r]arm-head-controller

knorth55:
clean up controller section.
normally, `arm-controller` should not contain `head-controller`.

I need to check with real robot.


knorth55:
@pazeshun what do you think?

knorth55:
I checked with real robot, and it works!

PR #2266, opened 2017-07-12
add controller-type in cancel-angle-vector

knorth55:
waiting for upstream https://github.com/jsk-ros-pkg/jsk_pr2eus/pull/313 

add `controller-type` in `:cancel-angle-vector`


wkentaro:
I tested on local pc.

knorth55:
It doesn't have error, but without upstream PR, cancel-angle-vector doesn't work properly.

PR #2276, opened 2017-07-13
Move gripper upward in :return-object to prevent collision

pazeshun:
To prevent collision between gripper and shelf or tote


pazeshun:
@wkentaro Please merge this

Issue #2210, opened 2017-07-04
graspingp for pinching is difficult for small objects


pazeshun:
![dsc_1075](https://user-images.githubusercontent.com/14994939/27812514-6f34b648-60aa-11e7-98f2-b9cec600b744.JPG)
We can use scale system to detect grasping as well as finger load and finger angle

pazeshun:
I'll postpone this

pazeshun:
> We can use scale system to detect grasping as well as finger load and finger angle

This is done with #2277 
Currently, graspingp for pinching is ignored in all cases, so we can avoid this issue.
Please reopen this when we need graspingp for pinching again.

PR #2277, opened 2017-07-13
Always set graspingp of pinching as true

pazeshun:
For #2210 
My idea is we can detect some object is picked by weight scales in `verify-object`.
So we can skip grasping checks of `:try-to-pinch-object` by making `:graspingp` return t when `:pinch` is set.


pazeshun:
Tested on real robot

pazeshun:
@wkentaro Please merge this.

PR #2288, opened 2017-07-14
Add look_around_bins experiment

wkentaro:
## look_around_bins

```bash
roslaunch jsk_arc2017_baxter baxter.launch moveit:=false
roslaunch jsk_arc2017_baxter stereo_astra_hand.launch
roslaunch jsk_arc2017_baxter look_around_bins_main.launch
```


wkentaro:
It does not look good, and we should have another solution.

PR #2298, opened 2017-07-16
calibrate right_hand_camera left camera

knorth55:


wkentaro:
Comparing with https://github.com/start-jsk/jsk_apc/pull/2295, which is better?

knorth55:
hmmm, it is hard to say. I use both calibration, but either are not totally correct.
But i thought this PR has better calibration...

PR #2286, opened 2017-07-14
try picking twice and not recognize

knorth55:
after suction picking failure, do not recognize `FCN` again and try pinch.
this is [0] of #2260

PR #2302, opened 2017-07-16
Add trial count and redesign state machine

knorth55:
merge after #2286 

# Main change
- add `:check-trial-fail-count`.
this is because `picking-fail-count` depends on arm, not target.
`picking-fail-count` is messed up in #2203, so I move separate these counters into two.

## pick task
pick_state_machine_faster
## stow task
stow_state_machine_faster


knorth55:
I tested with real robot.

wkentaro:
@knorth55 Please resolve the conflicts.

knorth55:
@wkentaro resolved conflicts and test passed!

PR #2303, opened 2017-07-17
Use pre-built docker image

wkentaro:


wkentaro:
I achieved 28min for testing.

image

PR #2309, opened 2017-07-18
Remove outlier in flex sensor values

wkentaro:
@pazeshun 

Maybe median filter is fine, so please refer this when you really need it.

image


pazeshun:
Thanks!
Perhaps https://github.com/start-jsk/jsk_apc/issues/2306#issuecomment-316059844 is true problem...

PR #2313, opened 2017-07-18
Calibrate camera parameter

YutoUchimi:
For left camera on both hands...
- updated IR intrinsic parameter
- updated extrinsic parameter between depth_optical_frame and rgb_optical_frame


YutoUchimi:
Probably this is still not good result except center region of frame...

PR #2317, opened 2017-07-19
Stabilize flex sensor

### LINKED ISSUE #2306: Stabilize flex sensor
曲げセンサのセンサ生値にはすれ値が混ざっているので、フィルタかけます

pazeshun:
Close #2306 

The value of flex sensor improves like below:

|Before this change|After this change|
|---|---|
|screenshot from 2017-07-18 20 20 45|screenshot from 2017-07-19 16 35 35|

## Changes
- Stop proximity sensor before measuring flex


pazeshun:
I've already written this new firmware to Arduino

PR #2323, opened 2017-07-19
increase weight error for pick task

knorth55:
increase weight error.
I tried pick task, and hose weight is around 10-20g.
so I increase error to 20g.
this is temporary solution, so I added FIXME comment.


pazeshun:
So what I should do is to make right hand hose short?

knorth55:
yes

pazeshun:
I made hose short.
Please test this

PR #2326, opened 2017-07-19
Fix error of weight_candidates_refiner for expo (20g)

wkentaro:


knorth55:
Because expo_eraser is 20g!

PR #2334, opened 2017-07-20
[WIP] Improve :stop-grasp

pazeshun:
## Changes
- Use `:wait-interpolation-until` in `:stop-grasp` to detect collision between the suction finger and environments (eg. objects in shelf or cardboard box)
- Update `*baxter*` after `:stop-grasp` to avoid unnecessary movement


pazeshun:
Changed to #2362

PR #2341, opened 2017-07-20
Set grasp style by graspability

knorth55:


knorth55:
I haven't tested with real robot

knorth55:
I tested with real robot and debugged.
It works!

knorth55:
NOTE: state machine updated
### pick task
pick_state_machine

### stow task
stow_state_machine

PR #2348, opened 2017-07-21
add object-in-hand as attached object scene

knorth55:
add object in hand as attatched object.
this PR avoid collision between object-in-hand and rack.


knorth55:
I tested with real robot.

PR #2355, opened 2017-07-21
Enable dynamic calib of prismatic joints

pazeshun:
To start calibrating a prismatic joint:
```
(send *ri* :start-prismatic-calib :rarm)
```
You can know whether a prismatic joint is in calibration:
```
(send *ri* :prismatic-calibratingp :rarm)  ;; t or nil returns
```

When the calibration finishes, the joint goes to the position commanded previously


pazeshun:
Close #2180

pazeshun:
Tested in real robot

PR #2351, opened 2017-07-21
Add rack leg scene and refine moveit scene operation

knorth55:
cardboard_rack_scene

- add rack leg scene to avoid collision.
-  refine moveit scene operation in main
   - now it takes 0.5 seconds, so I remove unnecessary moveit scene operations.
- calibrate `cardboard_marker.yaml`
- refine `:place-object` motion

PR #2357, opened 2017-07-21
update cardboard_marker yaml

knorth55:
I made cardboard c position lower in order to avoid collision.
also, i calibrated other cardboard position, too.


knorth55:
merged to #2351

PR #2362, opened 2017-07-22
Don't extend prismatic joint in :stop-grasp

pazeshun:
In order to avoid collision to other objects in `return-object` or cardboard boxes in `place-object`


pazeshun:
Tested in real robot

PR #2386, opened 2017-07-23
lift higher to avoid collision b/w obj adn shelf

knorth55:


knorth55:
collision is reported by @708yamaguchi

knorth55:
merged to #2389

PR #2389, opened 2017-07-23
lift object set :rotation-axis t to avoid collision

knorth55:
In order to avoid hard place pose, we reset pose when robot lift object.


knorth55:
I tested with real robot.

PR #2399, opened 2017-07-25
return-from :pick-object faster when graspingp nil

knorth55:
I haven't tested with real robot


knorth55:
I tested with real robot.

knorth55:
I tested with real robot in stow.

PR #2401, opened 2017-07-25
set less recognize trial times

knorth55:
I haven't tested with real robot.


knorth55:
I tested with real robot in stow

knorth55:
I tested with real robot in pick

PR #2402, opened 2017-07-25
set count-limit 5 for start-picking-fail-count

knorth55:
I haven't test this with real robot.


wkentaro:
Please resolve the conflict.

knorth55:
I tested with real robot in stow

knorth55:
I tested with real robot in pick

PR #2406, opened 2017-07-25
use stamp just after recognize pose for recognize

knorth55:


knorth55:
I tested with real robot in stow.

knorth55:
I tested with real robot in pick.

PR #2408, opened 2017-07-25
Resolve collision between fingers in fold-pose-back

pazeshun:
I haven't tested in real robot.


pazeshun:
Tested in real robot

wkentaro:
```
+ apt-get install -qq -y mesa-utils
WARNING: The following packages cannot be authenticated!
  mesa-utils
E: There are problems and -y was used without --force-yes
```

PR #2417, opened 2017-07-26
Remove unix::sleep in arc-interface.l

wkentaro:
Moved from https://github.com/start-jsk/jsk_apc/pull/2403


knorth55:
I checked with real robot in pick task

PR #2418, opened 2017-07-26
Return object randomly

wkentaro:
Moved from https://github.com/start-jsk/jsk_apc/pull/2394


knorth55:
I checked with pick task.

PR #2423, opened 2017-07-26
Abort picking when IK before grasping fails

pazeshun:


pazeshun:
Tested with real robot

PR #2426, opened 2017-07-26
modify return object in stow

708yamaguchi:


wkentaro:
Moved to https://github.com/start-jsk/jsk_apc/pull/2437

PR #2429, opened 2017-07-26
return-from when classifier timeout

knorth55:


wkentaro:
Moved to https://github.com/start-jsk/jsk_apc/pull/2420

PR #2471, opened 2017-07-28
Fix fast second suction

pazeshun:


wkentaro:
Moved to https://github.com/start-jsk/jsk_apc/pull/2473

PR #2500, opened 2017-08-01
Add speak interface to baxter-interface

wkentaro:


wkentaro:
I tested with the real robot.

PR #2512, opened 2017-08-03
Add :arc-reset-pose as faster reset-pose than :fold-pose-back

wkentaro:
image


knorth55:
Did you try this reset pose with pick task?
If there is no collision when placing pbject in cardboard c or when returning arm from cardboard a or b, it is better to use this pose.

wkentaro:
Yes.

PR #2523, opened 2017-08-04
Publish nan value when unstable by EkEwIDriver

wkentaro:
Close https://github.com/start-jsk/jsk_apc/issues/2529

-1 cannot represent unstable state of the scale, because the scale can output minus values.


wkentaro:
I tested with the Pick task.

pazeshun:
Is this temp fix until https://github.com/jsk-ros-pkg/jsk_recognition/pull/2160#issuecomment-320274814 is finished?

wkentaro:
I hit on this idea after commenting on jsk_recognition PR. But both is ok for me.

PR #2524, opened 2017-08-04
Add script to forcely recognize usbs

wkentaro:
Close https://github.com/start-jsk/jsk_apc/issues/2528

It diconnect and reconnect usbs by software.

```
rosrun jsk_arc2017_baxter force_recognize_usb_devices
```


wkentaro:
I tried to use Thinkpad P51 as the second external PC, but found that it does not solve our problem of missing usb devices.
So I decided to forcely recognize usb devices as I do so in this PR.

pazeshun:
What device did you try to move to external PC?
Perhaps we have to move more devices.

wkentaro:
4 scales.

PR #2563, opened 2017-08-10
Fix rotate wrist ik for :rarm (inverse of :larm)

wkentaro:
This solves problem reported at https://github.com/start-jsk/jsk_apc/issues/2560


wkentaro:
Because problem in #2560 occurs only with :rarm.

PR #2613, opened 2017-12-11
remove sleep in moveit methods

knorth55:
pr2eus_moveit uses service call, which do not need sleep.

minor changes:
- `:add-tote-scene` `:delete-tote-scene` support `:arms`
- refactoring


pazeshun:
> pr2eus_moveit uses service call, which do not need sleep.

New feature in https://github.com/jsk-ros-pkg/jsk_pr2eus/pull/324 ?

knorth55:
yes

PR #2612, opened 2017-12-09
Add left gripper_v7 model

pazeshun:
## What is added
* Mesh model of new gripper (left gripper_v7)
* URDF of Baxter with left gripper_v7 (baxterlgv7)
  * Header files and source file to describe tendon transmission (used by gripper controller)

screenshot from 2017-12-09 17 15 59
* Auto generation of Euslisp model of baxterlgv7
screenshot from 2017-12-09 17 29 44
  * Slightly lighter than current baxter model
     ```
     $ du -k baxter.l baxterlgv7.l 
     34508	baxter.l
     34380	baxterlgv7.l
     ```

PR #2617, opened 2018-01-09
Fix palm endpoint of left gripper_v7

pazeshun:
This fixes palm coordinate of left gripper_v7 (mistake in #2612 ).
- Palm coordinate in URDF
    - Before
screenshot from 2018-01-06 16 12 17
    - After
screenshot from 2018-01-06 16 15 00
- Palm coordinate in Euslisp (it's transformed from one in URDF for pinching) 
    - Before
screenshot from 2018-01-06 16 25 03
    - After
screenshot from 2018-01-06 16 30 19

### DIFF
--- a/jsk_arc2017_baxter/euslisp/lib/baxterlgv7.l
+++ b/jsk_arc2017_baxter/euslisp/lib/baxterlgv7.l
@@ -13,7 +13,7 @@
 (defclass jsk_arc2017_baxter::baxterlgv7-robot
   :super baxter-robot
   :slots (view-hand-pos- ik-prepared-poses- hand-joints-
-                         rarm-palm-endpoint-ik- larm-palm-endpoint-ik-))
+                         rarm-palm-end-coords- larm-palm-end-coords-))
 
 (defmethod jsk_arc2017_baxter::baxterlgv7-robot
   (:init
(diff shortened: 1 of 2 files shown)


knorth55:
Does this mean that `roseus` coordinates and `URDF` coordinates are different?

pazeshun:
Yes. I transform coordinate in https://github.com/start-jsk/jsk_apc/pull/2617/files#diff-c9c881a02c4a2def8db9105e466c0a40R64

pazeshun:
Same as gripper-v6 https://github.com/start-jsk/jsk_apc/blob/master/jsk_arc2017_baxter/euslisp/lib/baxter.l#L62

knorth55:
How about setting another coordinates instead of moving the coordinates?

pazeshun:
> How about setting another coordinates instead of moving the coordinates?

I just came up with the same idea as you...
Maybe I only have to change `:rarm-palm-endpoint` and `:larm-palm-endpoint`
I'll try.

pazeshun:
Setting other coordinates is done. I confirmed IK with `:move-palm-end t` solved correctly.

pazeshun:
@knorth55 Can you review this?
I'll make another PR to set other palm coodinates in baxter with gripper-v6 if you want.

jsk_arc2017_baxter/euslisp/lib/baxterlgv7.l
```diff
@@ -59,9 +60,17 @@
                    :left_gripper_l_finger_yaw_joint
                    :left_gripper_r_finger_roll_joint_a
                    :left_gripper_r_finger_roll_joint_b))
-    ;; translate palm end for good cylindrical grasp
-    (send self :rarm-palm-endpoint :translate #f(30 0 0) :local)
-    (send self :larm-palm-endpoint :translate #f(30 0 0) :local)
+    ;; set new palm end for good cylindrical grasp
+    (setq rarm-palm-endpoint-ik- (make-cascoords))
```
knorth55: `ik` -> `lk`?

pazeshun: Oh, sorry for confusing name...
I named them to show those are used in IK and different from original URDF link (right_gripper_palm_endpoint).
But now I think I should name them `rarm-palm-end-coords-` because those are coodinates, not link.
I'll fix this.

pazeshun: fixed

pazeshun:
Test passed! Please merge this

PR #2623, opened 2018-01-18
Fixes for demo on 2018-01-19

wkentaro:
- Don't ignore fcn32s.npz to remove it by git clean

for git clean

- Skip empty bin while visualizing JSON

To make visualization better.

- Fix unreasonable ~json_dir param to candidates_publisher

a bug introduced by https://github.com/start-jsk/jsk_apc/pull/2618

### DIFF
--- a/jsk_arc2017_common/node_scripts/candidates_publisher.py
+++ b/jsk_arc2017_common/node_scripts/candidates_publisher.py
@@ -20,8 +20,8 @@ class CandidatesPublisher(ConnectionBasedTransport):
             '~output/candidates', LabelArray, queue_size=1)
         self.srv = dynamic_reconfigure.server.Server(
             CandidatesPublisherConfig, self._config_cb)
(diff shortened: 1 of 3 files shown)


pazeshun:
@knorth55 Is it ok to fix candidates_publisher like this PR?
What you want to do in #2618 is accomplished?

jsk_arc2017_common/node_scripts/candidates_publisher.py
```diff
@@ -21,7 +21,7 @@ def __init__(self):
         self.srv = dynamic_reconfigure.server.Server(
             CandidatesPublisherConfig, self._config_cb)
         self.label_names = rospy.get_param('~label_names')
-        self.json_dir = rospy.get_param('~json_dir')
```
knorth55: `self.json_dir = rospy.get_param('~json_dir', None)`

knorth55:
thank you!

PR #2624, opened 2018-02-07
add robot-safe and collision-avoid-ik

knorth55:


knorth55:
Collision Avoidance Links
baxter_collision_apc

Collision Avoidance IK
baxter_collision_apc

knorth55:
use weight method for `:no-gripper` `:inverse-kinematics`

PR #2625, opened 2018-02-07
clean up ik->nearest-pose

knorth55:
I clean up codes.
Same IK was solved twice, which was bottleneck


knorth55:
I checked with real robot

PR #2627, opened 2018-02-13
Re-calibrate right vacuum pad tendon winder

pazeshun:


knorth55:
I confirmed with real robot!

PR #2630, opened 2018-04-04
Specify axis camera with hostname instead of ip address

pazeshun:
Because IP address sometimes changes.


pazeshun:
Currently, axis camera's ip is not `133.11.216.141`

PR #2636, opened 2018-05-17
Re-calibrate finger tendon winder

pazeshun:
Because the finger tendon pulley in right hand is chipped:
dsc_1837


wkentaro:
Rebased.

wkentaro:
Rebased.

PR #2637, opened 2018-05-20
Remove image_proc 1.12.20 from fc.rosinstall

wkentaro:
Now image_proc 1.12.23 is released.
http://repositories.ros.org/status_page/ros_indigo_default.html?q=image_proc


knorth55:
Good! thanks

PR #2650, opened 2018-07-29
Add if-blocks to use only the single arm

wkentaro:
For https://groups.google.com/a/jsk.imi.i.u-tokyo.ac.jp/forum/#!topic/baxter/dj4RfMT24PI

**Usage**

```bash
roslaunch jsk_arc2017_baxter baxter.launch launch_left:=false respawn_arduino:=false
```


wkentaro:
It's ok. Sorry for rushing you by the emails.
I was just surprised since the v8 gripper looks very similar to v6.

PR #2657, opened 2018-08-03
Make it slower to send avs for smoother motion (going to above target object)

wkentaro:
I found scale 5.0 is too fast.
(Maybe this needs the fact by a video.)


wkentaro:
It exceeds the limit of the maximum joint speed and the motion rattles. (がたがたする）

pazeshun:
I understand. Thank you.

PR #2661, opened 2018-09-20
Fix :recognize-object to return t or nil

pazeshun:
#2648 makes `is-recognized` time stamp or nil, but `:recognize-object` should return t or nil for state machine.

### DIFF
--- a/jsk_arc2017_baxter/euslisp/lib/pick-interface.l
+++ b/jsk_arc2017_baxter/euslisp/lib/pick-interface.l
@@ -143,7 +143,7 @@
                 (setq moving-object-p- t)
                 (send self :recognize-target-object arm :stamp (ros::time-now))))
             (incf recognition-count))))
-      is-recognized))
+      (not (null is-recognized))))
   (:return-from-recognize-object (arm)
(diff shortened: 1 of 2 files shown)


wkentaro:
Can we add `(not is-recognized)` in pick-main or stow-main.l?

pazeshun:
> Can we add `(not is-recognized)` in pick-main or stow-main.l?
> Some of my programs need `is-recognized` to be timestamp.

I created #2662 to fulfil your request.
However, I'm against your idea because all methods in `pick-interface` or `stow-interface` called from `pick-main` or `stow-main` return t or nil.
#2662 violates that regularity.

wkentaro:
I'm not sure why we need to set that regularity. Do you have any idea? @knorth55 
Maybe we can add tests for that.

pazeshun:
Oops, my misunderstanding...
https://github.com/start-jsk/jsk_apc/blob/master/jsk_arc2017_baxter/euslisp/pick-main.l#L81-L82
Now I allow #2662

pazeshun:
Closed
https://github.com/start-jsk/jsk_apc/pull/2662#issuecomment-423604708

PR #2664, opened 2018-10-16
Try docker image with latest apt pkgs

pazeshun:
This PR checks if new docker image solves travis failure.

I created a container from docker image `wkentaro/jsk_apc:<indigo/kinetic>`, executed `apt-get update && apt-get dist-upgrade` in that container, and committed that container to docker image `pazeshun/jsk_apc:<indigo/kinetic>`.

If travis test of this PR passes, we just need to update `wkentaro/jsk_apc:<indigo/kinetic>` manually to pass travis test from now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ env:
     - ROS_DISTRO=kinetic
 
 script:
-  - export DOCKER_IMAGE="wkentaro/jsk_apc:$ROS_DISTRO"
+  - export DOCKER_IMAGE="pazeshun/jsk_apc:$ROS_DISTRO"
   # build & test ROS packages
   - source .travis/travis.sh
   # build doc


wkentaro:
I see. Please apply the same change to below though I'm not sure adding dist-upgrade is a standard way. Could you add `FIXME: https://github.com/start-jsk/jsk_apc/pull/2664` when you change the Dockerfile?

- https://github.com/start-jsk/jsk_apc/blob/9628a7deb86728012865179ac1e14f3d84f3d0a5/docker/indigo/Dockerfile#L7
- https://github.com/start-jsk/jsk_apc/blob/9628a7deb86728012865179ac1e14f3d84f3d0a5/docker/kinetic/Dockerfile#L7

wkentaro:
Closed via #2665

PR #2665, opened 2018-10-16
Add dist-upgrade to Dockerfiles

pazeshun:
From https://github.com/start-jsk/jsk_apc/pull/2664#issuecomment-430179884


wkentaro:
Yeah. Thanks!

PR #2671, opened 2018-12-16
add baxter_moveit_config as run_depend

knorth55:


pazeshun:
Can you also add this to jsk_2016_01_baxter_apc?
https://github.com/start-jsk/jsk_apc/blob/master/jsk_2016_01_baxter_apc/launch/include/ompl_planning_pipeline.launch#L20

knorth55:
Updated

PR #2676, opened 2019-04-03
Add check arduino node

knorth55:
- add check_arduino node
  - if no topic is published but node is running, kill the node


knorth55:
we test this node in 20190405 demo.

knorth55:
I tested this node but the arduino error does not happen even when the robot uses dual arm grasping...

PR #2683, opened 2019-05-13
add dynamixel_controllers as build_depend

knorth55:
add `dynamixel_controllers` as `build_depend`


pazeshun:
I think only run_depend is required.
Why is build_depend is also required?

knorth55:
because you include dynamixel_controllers/EnableTorque.h  in gripper_v6_controller.cpp

pazeshun:
> because you include dynamixel_controllers/EnableTorque.h in gripper_v6_controller.cpp

I've totally forgot. I'm very sorry...

PR #2702, opened 2019-11-02
use VOC dataset in google drive

knorth55:
official website seems down ;(
This PR fixes Travis build failure.


knorth55:
VOC official site is now working but it seems unstable.
should we switch to our own google drive?

pazeshun:
> should we switch to our own google drive?

I feel this is better, because I don't trust official website much.
However, repeating this will make what we manage too much...
@knorth55 What's your opinion? If you don't trust official one, either, let's switch to our google drive.

wkentaro:
This is only for the test, right? Maybe we should use more trustable test files: e.g., https://www.thinkbroadband.com/download

wkentaro:
Sent PR https://github.com/start-jsk/jsk_apc/pull/2703

PR #2703, opened 2019-11-03
Stop using VOCdevkit tar file to test extractall function

wkentaro:
Close https://github.com/start-jsk/jsk_apc/pull/2702


knorth55:
LGTM
BTW, are you using VOC data in occlsegm or grasp fusion?
if so, we should  switch to google drive link to download it.

wkentaro:
This is devkit, not dataset, so should be fine, right?

wkentaro:
I found that 5MB.zip is not actually the zip file.

knorth55:
LGTM. I prefer using `hub` released tar.gz.

PR #2705, opened 2020-01-17
Change FA-I-sensor to root repository in *.rosinstall

pazeshun:
Because https://github.com/RoboticMaterials/FA-I-sensor/pull/11 is merged

### DIFF
--- a/.travis.rosinstall
+++ b/.travis.rosinstall
@@ -1,7 +1,7 @@
 - git:
     local-name: FA-I-sensor/force_proximity_ros
-    uri: https://github.com/knorth55/FA-I-sensor.git
-    version: jsk_apc
+    uri: https://github.com/RoboticMaterials/FA-I-sensor.git
+    version: master
 - git:
     local-name: RethinkRobotics/baxter_common
     uri: https://github.com/knorth55/baxter_common.git
--- a/fc.rosinstall.indigo
+++ b/fc.rosinstall.indigo
@@ -44,8 +44,8 @@
     version: gripper-v6-devel
 - git:
     local-name: FA-I-sensor/force_proximity_ros
-    uri: https://github.com/knorth55/FA-I-sensor.git
-    version: jsk_apc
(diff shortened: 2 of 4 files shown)


knorth55:
travis failed because of setuptools
this PR fixes this problem https://github.com/locusrobotics/catkin_virtualenv/pull/49

pazeshun:
@knorth55 Thank you.
Just FYI:
This setuptools problem seems due to some PyPA mirrors which wrongly drop `python_requires = >=3.5` setting:
https://github.com/pypa/virtualenv/issues/1493#issuecomment-573461695
https://github.com/pypa/pip/issues/7586#issuecomment-573460206

knorth55:
so, we need to wait for a while and the problem will be solved?

pazeshun:
> so, we need to wait for a while and the problem will be solved?

We just have to wait for https://github.com/locusrobotics/catkin_virtualenv/pull/49 to be released.
The wrong mirrors don't seem fixed yet, so https://github.com/locusrobotics/catkin_virtualenv/pull/49 explicitly install `setuptools<45` to stop downloading setuptools 45.0.0.

pazeshun:
@knorth55 
BTW, will chainer problem on Python 2 be fixed?
I think https://github.com/chainer/chainer/pull/8517 should be included in chainer 7.0.0, otherwise 7.0.0 remains to be installed on Python 2.
However, I couldn't find `python_requires` in https://github.com/chainer/chainer/blob/v7.0.0/setup.py and https://github.com/chainer/chainer/blob/v7.0.0/setup.cfg.

knorth55:
I hoped so, but they released in chainer v7.1.0

pazeshun:
I commented the issue: https://github.com/chainer/chainer/pull/8517#issuecomment-576128957

knorth55:
@pazeshun can I merge this? the chainer problem does not seem to be solved and it is not related to this PR.

pazeshun:
@knorth55 Yes, please merge this.

PR #2706, opened 2020-01-23
[jsk_arc2017_baxter] Set png_level of compressedDepth to keep image_raw hz on kinetic

pazeshun:
When subscribing compressedDepth on kinetic, hz of image_raw drops.
This is because load of CPU processing camera becomes too high.
To avoid this on baxter-c1, png_level should be under 6.

This behavior is different from indigo, so we should do further investigation...
I created https://github.com/ros-perception/image_common/issues/151 to gather information.
cc. @YutoUchimi


pazeshun:
This PR is already tested with real robot.

knorth55:
@pazeshun can i merge this?

pazeshun:
@knorth55 Yes, please merge this.

PR #2712, opened 2020-05-12
Stop downloading unused model checkpoints

wkentaro:
For https://github.com/start-jsk/jsk_apc/pull/2707#issuecomment-627229028


pazeshun:
Whole cache size is reduced to 18085.82MB, which means number of downloaded files seems correctly reduced:
jsk_apc_cache
Now I'll restart travis to use uploaded cache.

PR #2722, opened 2020-09-09
add sleep in calib_required_controller

knorth55:
@pazeshun 
we found this error with softhand.
original commit is here https://github.com/knorth55/softhand_ros/commit/e196a45ea0d25894f1c09bcbdcc42942abf113f4


pazeshun:
We will continuously check if this is really required through @knorth55 's use case.

knorth55:
can you review agian this PR?
I will merge this if OK.

pazeshun:
Sorry, could you add the same change to https://github.com/start-jsk/jsk_apc/blob/master/demos/sphand_ros/sphand_driver/src/sphand_driver/calib_required_joint_controller.py?

knorth55:
I added

pazeshun (approved):
Thank you!

PR #2726, opened 2020-10-17
add grasp_data_generator for IROS2018 and AdvancedRobotics2020

knorth55:
I add dataset generation and model training code for my research.


knorth55:
@pazeshun it is hard to review, so i will merge this if you are ok. this scripts does not affect to jsk_apc main programs.

pazeshun (approved):

PR #2739, opened 2021-05-20
change to pure_nccl: hierarchical is no more available

knorth55:
https://docs.chainer.org/en/stable/chainermn/reference/index.html#chainermn.create_communicator

`hierarchical` is no more available...


pazeshun:
Cannot we specify chainermn version?

knorth55:
chainermn is optional dependency.
we can inform that we should install chainermn in lower version, but hierarchical is not recommended in lower version, too.

pazeshun (approved):
I see, thanks

PR #2747, opened 2021-06-28
copy rbo and stop using submodule

knorth55:
maybe we should stop using personal repo as submodule.


.gitmodules
```diff
@@ -1,7 +1,3 @@
-[submodule "jsk_apc2016_common/python/jsk_apc2016_common/rbo_segmentation"]
-	path = jsk_apc2016_common/python/jsk_apc2016_common/rbo_segmentation
-	url = https://github.com/yuyu2172/rbo-apc-object-segmentation
- 	branch = jsk_apc
```
pazeshun, line 4: This says we uses `jsk_apc` branch.
However, I think you copied `apc2016` branch.
Which is correct?

pazeshun: I understand #1776 forgot to change `.gitmodules` and `.gitmodules` is wrong.

pazeshun (approved):

knorth55:
You need to deinit first for pulling master
```
git submodule deinit --all
git pull origin master
git submodule init
git submodule update
```

PR #2748, opened 2021-06-30
install moveit_robots from source

knorth55:
latest moveit removes move_group/MoveGroupExecuteService

we need these commits.
https://github.com/ros-planning/moveit_robots/commit/707b86b43ba2a0bda1d27ce2019f798d9d79005a
https://github.com/ros-planning/moveit_robots/commit/3514d389b45361b19af936099af45090912ec1f1


pazeshun:
Don't we have to change `fc.rosinstall.kinetic` as well?

knorth55:
no because released moveit in kinetic still have the module

pazeshun (approved):
Thanks

PR #2770, opened 2023-12-21
[sphand_driver] Add arduino sketch to test WrPPS Single Board

pazeshun:


pazeshun:
Tested with a real sensor board.

